PR #50, opened 2022-05-20
chore(deps): upgrade designer to fl_chart 0.5

johannesvedder:


nstrelow:
Nice, great that you migrated it.

Were you able to test it?

johannesvedder:
It is working, but I wasn't 100% satisfied with the current state, since there were still some visual margin and padding glitches I couldn't fix. It might have improved with fl_chart 0.51 introducing a new `space` property. I'll figure that out.

PR #51, opened 2022-05-20
chore(deps): upgrade fhir

johannesvedder:


nstrelow:
Sadly this is currently breaking the not so great FHIR questionnaire support we have. Would keep off until somebody has time to fix it.

johannesvedder:
https://github.com/hpi-studyu/studyu/pull/23#issuecomment-1128167653 as a reference for what to work on.

PR #53, opened 2022-05-20
chore: github workflow for publishing to pub.dev

johannesvedder:
https://github.com/marketplace/actions/dart-and-flutter-package-publisher

### DIFF
--- a/.github/workflows/publish_pubdev.yml
+++ b/.github/workflows/publish_pubdev.yml
@@ -23,20 +23,20 @@ jobs:
       - name: 'Checkout'
         uses: actions/checkout@v2 # required!
 
-      - name: '>> Dart package app <<'
+      - name: '>> Dart package core <<'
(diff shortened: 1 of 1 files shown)


nstrelow:
I don't think we want to publish the app and designer package.

Also this is not using melos, so it cannot figure out in which order to upload the dependencies. 

But I think you can adjust the order of steps so it always first uploads:
1. core
2. flutter_common

nstrelow:
Oh, and this seems totally fine and a great idea. Hope it works.

I think you should be able to do all the required steps now by yourself, but let me know if this is not the case.

nstrelow (approved):
Looks great if you added the secrets

johannesvedder:
Thanks for the feedback. I'll leave the `dryRunOnly` field active for the next test run and remove it if everything works fine.

PR #64, opened 2022-06-10
Refactoring designer_v2 to Riverpod

johannesvedder:
Refactoring designer_v2 to use the [Riverpod ](https://riverpod.dev/de/) state management.

## TODO

Files to refactor:

### General

- [x] main
- [x] router
- [ ] app_state
- [ ] user (needs to be removed)

### Pages

- [x] my_app (mostly finished)
- [ ] error_page
- [x] login_page
- [ ] splash_page
- [x] studyu_dashboard_screen


johannesvedder:
We probably need access to firebase to add the designer-v2 there to make the preview work.

nstrelow:
Very cool, much success with the refactoring :D

janukobytsch:
Thanks for taking a look at Riverpod @johannesvedder !

After reading through the articles you sent & reviewing some examples, I'm not sure yet that it's the best approach for the project moving forward. A few issues I see:
- There's a steep learning curve and the documentation is really poor. This will be a big challenge for any student team working on StudyU in the future
- Little value-add over the standard provider API. Tbh I'm not even sure we have any of the problems that Riverpod is trying to solve (simplifying dependencies between providers, shadowing of providers of the same type at different layers in the widget hierarchy, avoiding runtime errors due to resolution by type). Compile-time safety for Provider.of<T> calls is nice-to-have, but I don't think it will be a huge issue for us.
- Maybe I still don't understand Riverpod, but it doesn't seem like a real state management solution to me in the sense that it would be opinionated about how to structure the state & how to architect the app. The way I see it it's just a mechanism for providing objects anywhere in the code while avoiding global scope issues (because the state is boxed/scoped transparently) - so it's primarily about dependency injection. Yes, you can also use it to construct observer graphs & derive computed state (e.g. fetching a list of entities in a FutureProvider & then filtering the list in a downstream provider by some criterion), but do we want to go that direction? Personally I'd prefer having a List<Study> variable somewhere that I can manipulate along with some business logic rather than a FutureProvider that I don't have control over & that lives in a different place than my business logic. Chaining providers to derive / compute state reminds me too much of Redux' selectors, it's also very different from the StateNotifier/ChangeNotifier approach (also suggested by Riverpod). I'm a bit stunned at the level of flexibility & different architectures you could build on top of Riverpod. To be honest I think that level of flexibility & lack of opinion makes the library easy to misuse, overly complex & hard to get into.
- Did you find a complete & well-documented / explained example project for Riverpod that goes beyond the counter app with actual business logic + data fetching + deserialization? This [[article](https://blog.codemagic.io/flutter-state-management-with-riverpod/)](https://blog.codemagic.io/flutter-state-management-with-riverpod/) kind of explains it, but doesn't have any business logic + I find it very weird that they have an "HomePageInitial" widget just for initiating the network request.

Let me know what you think about this & the comments I added. My feeling is that Riverpod is a bit overkill right now & it would be simpler to just manage everything with Provider (dependency injection) and ChangeNotifier (state management) - it's also super easy to create a ViewModel-type architecture with these two. If we need something more advanced, we can always bring in Riverpod later. If we properly separate our widgets into presentational & connected**, migrating to a different state management should be even easier. If we have good reasons for sticking with Riverpod, we definitely need some clear guidelines on how to use it (and maybe just use it as a drop-in replacement for provider without any of the observer graph / derived state stuff).

** For inspiration: In Redux architectures there is a distinction between dumb/presentational components and connected components. Dumb components are just stateless widget and the connected components are a wrapper that reads+subscribes to redux stores/selectors & passes in the data as props to the wrapped widget. I feel like this would be more elegant to decouple the injection/subscription from the actual UI than having ref.read/watch scattered across the build method. Makes it pretty easy to see all data dependencies in a single place + change to a different state management solution if needed. The downside is there is some boilerplate because you need to create a wrapper for each widget that has a data dependency.

janukobytsch:
@nstrelow Have you looked into Riverpod before or any other state management approaches? We definitely want something more solid & testable than what's there currently ("AppState" ChangeNotifier singleton + business logic/networking tightly coupled to UI widgets)

PR #79, opened 2022-06-20
Implement dashboard (Riverpod architecture)

janukobytsch:
For early review

Remaining TODOs (before this is ready to be merged):
- [x] Migrate authentication & routing to Riverpod (re-integrate & clean up changes from other branch)
- [x] Implement study filters in DashboardState based on user auth status
- [x] Replace usage of MockedApiClient with SupabaseObject methods


johannesvedder:
I have added an early version of a login page made with flutterflow for testing purposes, together with a rewrite of the authentication code for the use with Riverpod. Supabase connection for login/logout is also working. The changes of the old branch related to Riverpod are also merged, but there is still room for improvement on the architecture I would say. We can have a talk about that in the next days @janukobytsch :)

janukobytsch:
@johannesvedder Cleared all outstanding TODOs, should be good to merge now - please have a look :)

designer_v2/lib/features/dashboard/dashboard_state.dart
```diff
@@ -0,0 +1,52 @@
...
+  /// Currently selected filter to be applied to the list of studies
+  /// in order to determine the [visibleStudies]
+  final StudiesFilter studiesFilter;
+
+  /// Currently authenticated user (used for filtering studies)
+  final User currentUser;
+
+  /// The currently visible list of studies as by the selected filter
+  List<Study> get visibleStudies =>
+      studiesFilter.apply(studies: studies, user: currentUser).toList();
```
johannesvedder, line 32: Shouldn't `visibleStudies` be equal to `all` studies?

designer_v2/pubspec.yaml
```diff
@@ -9,19 +9,36 @@ environment:
...
     sdk: flutter
-  flutter_lints: ^2.0.0
+  flutter_lints: ^2.0.1
 
 flutter:
   uses-material-design: true
-
+  # todo this can be removed if there are no more local asseets
+  assets:
+    - assets/data/
```
johannesvedder, line 44: We can probably remove or disable this as long as we don't need it, since it generates a lot of warnings that the directory cannot be found.

designer_v2/lib/router.dart
```diff
@@ -1,82 +1,152 @@
...
-  splash(title: "splash", path: "/splash"),
-  login(title: "login", path: "/login"),
-  error(title: "error", path: "/error");
+  dashboard(id: "dashboard", path: "/studies"),
+  dashboardOwned(id: "dashboardOwned", path: "owned"),
+  dashboardShared(id: "dashboardShared", path: "shared"),
+  studyRegistry(id: "studyRegistry", path: "registry"),
+  splash(id: "splash", path: "/splash"),
+  login(id: "login", path: "/login"),
+  error(id: "error", path: "/error");
```
johannesvedder, line 24: This can become pretty messy with time. Maybe we can find a way to structure the paths more hierarchically?

designer_v2/lib/router.dart
```diff
@@ -1,82 +1,152 @@
...
+        name: RouterPage.dashboardShared.id,
+        builder: (context, state) => DashboardScreen(),
       ),
       GoRoute(
-        path: RouterPage.login.path,
-        name: RouterPage.login.title,
-        builder: (context, state) => LoginPage(),
+        path: RouterPage.studyRegistry.path,
+        name: RouterPage.studyRegistry.id,
+        builder: (context, state) => DashboardScreen(),
```
johannesvedder, line 52: Same as above. Would be great to have some kind of hierachy to redirect everything part of the dashboard to `DashboardScreen()`

johannesvedder (approved):
Looks great overall! 👍 Just some small notes from my side.

designer_v2/lib/features/dashboard/dashboard_state.dart
janukobytsch: Only if `studyFilter == StudyFilter.all` because the filter is a no-op in that case (see StudyFilter, the logic here is copied from the old designer). This is in fact how the public Study Registry page is implemented currently (not ideal for obvious reasons).

Normally `visibleStudies` is a subset of all studies. The full set of studies that the user can access & that is returned from querying the Studies table in Supabase is stored in `DashboardState.studies` (fetched & synced from the repository stream). The `studiesFilter` is controlled by the current dashboard route (e.g. navigating to "Shared Studies" will set the `studiesFilter` to `StudyFilter.shared`). Because `studiesFilter` is used in the state's equality check (see `Equatable.props`), updating the state with the new `studiesFilter` will result in a rerender of the widgets that are watching the `DashboardController` (this is just how StateNotifier works in Riverpod to enable reactivity). When the widget is rerendered, it calls the `visibleStudies` getter (dynamically applying the new filter to the same set of all studies that are accessible to the user). This way we can reuse the same widget to display a different subset of studies on different routes (/studies/owned, /studies/shared, /studies/registry).

designer_v2/lib/router.dart
janukobytsch: Agreed, but not sure how to do this if we want to keep using an Enum here. I would not worry about it for now until it becomes a problem.

designer_v2/lib/router.dart
janukobytsch: It's not clear to me what problem you are describing here.

Is it the duplication of three different routes all rendering the same widget? Unless you want to write a custom parser to enable more concise route configuration, there's not really a way around that because every route & the corresponding widget need to be registered at some point.

PR #122, opened 2022-07-14
fix: JSON Parse Error in Study Result

johannesvedder:
Fixes a bug that crashes the app when deserializing a List of Strings which are parsed incorrectly as dynamic objects.


janukobytsch:
@johannesvedder please make sure this won't break when merging

PR #92, opened 2022-07-05
feat(designer_v2): Study Preview

johannesvedder:
Adds a study preview function to the `Test` screen.

Depends on https://github.com/hpi-studyu/studyu/pull/122

PR #123, opened 2022-07-19
Feat: Confirmation modal for destructive actions

janukobytsch:
Depends on #92


johannesvedder (approved):
LGTM! The modal on clicking delete shows up and the notification when accepting the alert works as expected. 🎉

janukobytsch:
@johannesvedder I have removed the dependency on #92 & re-opened this PR in #148 . You should be able to use it for your updated survey preview now.

PR #128, opened 2022-07-25
feat: Recruit tab / Invite Code

janukobytsch:
Depends on #123


johannesvedder:
Doesn't seem to work for me. When creating a new study and the design tab opens, the page is empty and multiple errors occur. Most promising one:

```
======== Exception caught by rendering library =====================================================
The following assertion was thrown during performLayout():
RenderPadding object was given an infinite size during layout.

This probably means that it is a render object that tries to be as big as possible, but it was put inside another render object that allows its children to pick their own size.
The nearest ancestor providing an unbounded height constraint is: _RenderSingleChildViewport#35893 relayoutBoundary=up20 NEEDS-LAYOUT NEEDS-PAINT NEEDS-COMPOSITING-BITS-UPDATE
...  needs compositing
...  parentData: <none> (can use size)
...  constraints: BoxConstraints(w=1536.0, 0.0<=h<=697.6)
...  size: MISSING
...  offset: Offset(0.0, -0.0)
The constraints that applied to the RenderPadding were: BoxConstraints(w=1536.0, 0.0<=h<=Infinity)
The exact size it was given was: Size(1536.0, Infinity)
```

Full stacktrace: https://pastebin.com/S7wv0uHB

I cleaned my workspace and ran bootstrap again, but still no luck. 🤔

janukobytsch:
@johannesvedder reopened this PR as #149 with a quickfix to get the legacy designer working again

PR #381, opened 2023-05-25
fix/gitignore-build-dirs

jannis-baum:


.gitignore
```diff
@@ -80,4 +80,4 @@ generated/
 .firebase
 
 # Ruby bundler files for fastlane
-.bundle
\ No newline at end of file
+.bundle
```
johannesvedder: Let's stick to newlines at the end of files.

johannesvedder:
Good catch. We should consider writing some test cases at some point. :D

johannesvedder (approved):

.gitignore
jannis-baum: Totally agree😇

PR #454, opened 2023-06-15
Remove legacy FHIR support

### LINKED ISSUE #453: Remove legacy FHIR support

jannis-baum:
Close #453

Created [tag](https://github.com/hpi-studyu/studyu/releases/tag/legacy%2Ffhir-support) in case anyone still needs this in the future


johannesvedder:
We would probably also want to remove this from the sql schema: https://github.com/hpi-studyu/studyu/blob/6a6b9925ba30fe49b238014cc265a648357078f3/database/studyu-schema.sql#L92

jannis-baum:
> We would probably also want to remove this from the sql schema:
> 
> https://github.com/hpi-studyu/studyu/blob/6a6b9925ba30fe49b238014cc265a648357078f3/database/studyu-schema.sql#L92

Is this safe to just remove?

johannesvedder:
And let's discuss the changes in our next meeting first, because this is a breaking change

johannesvedder:
> > We would probably also want to remove this from the sql schema:
> > https://github.com/hpi-studyu/studyu/blob/6a6b9925ba30fe49b238014cc265a648357078f3/database/studyu-schema.sql#L92
> 
> Is this safe to just remove?

This field is only used for fhir questionaires, so if we decide to drop supporting fhir we should also remove it from the database

johannesvedder:
PR https://github.com/hpi-studyu/studyu/pull/51 could live on as a long-term issue if we would like to include support in the future at some point

jannis-baum:
> This field is only used for fhir questionaires, so if we decide to drop supporting fhir we should also remove it from the database

I agree! But what about the existing database; do we have to migrate something manually or can we just remove it from the schema?

johannesvedder:
> > This field is only used for fhir questionaires, so if we decide to drop supporting fhir we should also remove it from the database
> 
> I agree! But what about the existing database; do we have to migrate something manually or can we just remove it from the schema?

If there are no entries in the column that means no one is using it and we can just delete the column. No migration needed :)

johannesvedder (approved):

johannesvedder (approved):

Issue #455, opened 2023-06-15
chore: Refactor Github Workflows


johannesvedder:
- At the start of any workflow depending on the studyu packages on pub.dev it is checked if the `Publish to Pub.dev` workflow is running. Only if this is the case, the run is delayed with a timeout of some minutes (estimated guess of max duration that publishing takes). Otherwise, the job should start directly

PR #456, opened 2023-06-15
refactor(#455): github workflow

johannesvedder:


johannesvedder:
Usw för all workflow without melos except pubdev:

https://stackoverflow.com/questions/62750603/github-actions-trigger-another-action-after-one-action-is-completed

PR #457, opened 2023-06-17
feat: Designer Dashboard Improvements

### LINKED ISSUE #432: Feature(designerV2): Dashboard Improvements
[Trello card](https://trello.com/c/kbYf9E01)

johannesvedder:
closes #432

- [x] Search
- [x] Filter
- [x] Pin studies to the top

## More todos moved to issue https://github.com/hpi-studyu/studyu/pull/494

- Add Tags
- Click on tags filters studies. Click again deselects
- Clean up all new todos


designer_v2/lib/features/dashboard/dashboard_state.dart
```diff
@@ -25,33 +26,64 @@ class DashboardState extends Equatable {
...
+        if (pinnedStudies.contains(study.id)) {
+          pinned.add(study);
+          return true;
+        }
+        return false;
+      });
+
+      // Insert pinned studies at the beginning of the filteredStudies list
+      sortedStudies.insertAll(0, pinned);
+    }
```
jannis-baum, line 72: Would it be possible to simplify this by just adding the `is pinned?`-condition to the sorting predicate? I.e. if `pinnedStudies.contains(other) - pinnedStudies.contains(study)` is not 0 we return this difference and else the current predicate

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -14,60 +17,146 @@ class StudiesTable extends StatelessWidget {
...
+  List<int Function(Study a, Study b)?> get _sortColumns {
+    final predicates = [
+      (Study a, Study b) => 0, // do not sort pin icon
+      (Study a, Study b) => a.title!.compareTo(b.title!),
+      (Study a, Study b) => a.status.index.compareTo(b.status.index),
+      (Study a, Study b) => a.participation.index.compareTo(b.participation.index),
+      (Study a, Study b) => a.createdAt!.compareTo(b.createdAt!),
+      (Study a, Study b) => a.participantCount.compareTo(b.participantCount),
+      (Study a, Study b) => a.activeSubjectCount.compareTo(b.activeSubjectCount),
+      (Study a, Study b) => a.endedCount.compareTo(b.endedCount),
```
jannis-baum, line 114: Not sure if this is the right place for this or if it happens somewhere else/is already taken care of:

Since Dart's `sort` is not stable for equally comparing items, maybe we should sort by IDs as a final fallback so nobody gets confused that the order can randomly change for studies that compare to be equal?

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -14,60 +17,146 @@ class StudiesTable extends StatelessWidget {
...
+      final bool contains = pinnedStudies.contains(item.id);
+      final bool hovers = state.contains(MaterialState.hovered);
+      switch (hovers) {
+        case true:
+          return contains ? icon(MdiIcons.pinOff) : icon(MdiIcons.pin);
+        case false:
+          return contains ? icon(MdiIcons.pin) : const SizedBox.shrink();
+        default:
+          return const SizedBox.shrink();
+      }
```
jannis-baum, line 144: Can the default case even happen? Why not

```suggestion
      if (hovers) {
        return contains ? icon(MdiIcons.pinOff) : icon(MdiIcons.pin);
      } else {
        return contains ? icon(MdiIcons.pin) : const SizedBox.shrink();
      }
```

designer_v2/lib/features/design/study_form_controller.dart
```diff
@@ -151,7 +151,7 @@ class StudyFormViewModel extends FormViewModel<Study> implements IFormViewModelD
   Future _applyAndSaveSubform(IStudyFormData subformData) {
     studyDirtyCopy ??= formData!.exactDuplicate();
     subformData.apply(studyDirtyCopy!);
-    // Flush the on-write study copy to the repository & clear it
+    // Flush the on-write study copy to the repository and clear it
```
jannis-baum, line 154: There are quite a few other places with `&` in comments; if this is relevant we should consider changing all of them

designer_v2/lib/localization/app_de.arb
```diff
@@ -193,11 +198,16 @@
   "form_field_study_title_required": "Die Studie muss einen Titel haben",
   "form_field_study_title_default": "Unbenannte Studie",
   "form_field_study_description": "Beschreibung",
-  "form_field_study_description_tooltip": "Beschreibe die Studie kurz & verständlich für Teilnehmer",
-  "form_field_study_description_hint": "Beschreibe die Studie kurz & verständlich für Teilnehmer",
+  "form_field_study_description_tooltip": "Beschreibe die Studie kurz und verständlich für Teilnehmer",
+  "form_field_study_description_hint": "Beschreibe die Studie kurz und verständlich für Teilnehmer",
```
jannis-baum, line 202: The arb files also still have other ampersands; would also consider changing all of them if they were a problem here

jannis-baum (commented):
Looks great! See little comments & some thoughts on UI/UX below

---

### UI/UX (feel free to object!)

- the search bar has this skeuomorphic design that afaik is not used anywhere else in the designer; i would switch to the flat style to make it look more streamlined
- the ability to disable/unclick the sorting to me only makes sense if there is a direct use for the order the studies are shown in in this state. right now it looks like it's just whatever order they have in the database which the user shouldn't have to care about (unless this does have a meaning, then this meaning should be communicated through a clickable field as well). i would do one of
  1. disable the "don't sort" option by always sorting by something and displaying this accordingly with the arrow (e.g. by default/initially we sort by name descending). clicking one of the fields then just toggles the arrow back and forth and there is no way to completely get rid of it
  2. give having all arrows disabled the meaning of sorting by a custom order (compare spotify playlists). with this the user should be able to drag & drop to reorder the studies. then the default/initial state can be that no arrows are selected and this custom order is shown. dragging a study when any arrow is disabled (i.e. sorting by anything but the custom order) should not be allowed

designer_v2/lib/features/dashboard/studies_table.dart
johannesvedder: This did not occur during my tests, but I think it's worth implementing the addition. 👍

jannis-baum (approved):

PR #470, opened 2023-07-01
Docker Volumes For Supabase Deployment

florian-papsdorf:
Currently, the Supabase backend is deployed with docker-compose whereby, all persistent data is stored in directories on the host machine. 
Docker provides a convenient way to persist data in so-called volumes. Thereby, we do not have to take care about where the productive data lives inside the file tree of the repository during updating the deployment. We introduce these volumes for storing all postgres and Supabase blob storage data.

### DIFF
--- a/docker/supabase/docker-compose-db.yml
+++ b/docker/supabase/docker-compose-db.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.8'
 name: 'studyu'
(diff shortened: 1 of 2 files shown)


johannesvedder:
Thanks! I have made some changes to follow our conventional commit guideline and added some documentation in favor of #471

PR #475, opened 2023-07-04
Feat/docker improvements

johannesvedder:
- Some modifications to the docker setup eliminating warnings, renaming files and containers
- Upgrade the setup from the original Supabase repo.
- Prepare the .env files for the new HPI setup


flutter_common/lib/src/utils/env_loader.dart
```diff
@@ -13,8 +13,13 @@ String envFilePath() {
...
   await dotenv.load(fileName: envFilePath());
-  final supabaseUrl = dotenv.env['STUDYU_SUPABASE_URL']!;
-  final supabaseAnonKey = dotenv.env['STUDYU_SUPABASE_PUBLIC_ANON_KEY']!;
+  final supabaseUrl = dotenv.env['STUDYU_SUPABASE_URL'] ?? const String.fromEnvironment('STUDYU_SUPABASE_URL');
+  final supabaseAnonKey =
+      dotenv.env['STUDYU_SUPABASE_PUBLIC_ANON_KEY'] ?? const String.fromEnvironment('STUDYU_SUPABASE_PUBLIC_ANON_KEY');
+  final envAppUrl = dotenv.env['STUDYU_APP_URL'] ?? const String.fromEnvironment('STUDYU_APP_URL');
+  final envDesignerUrl = dotenv.env['STUDYU_DESIGNER_URL'] ?? const String.fromEnvironment('STUDYU_DESIGNER_URL');
+  final envProjectGeneratorUrl =
+      dotenv.env['STUDYU_PROJECT_GENERATOR_URL'] ?? const String.fromEnvironment('STUDYU_PROJECT_GENERATOR_URL');
```
jannis-baum, line 22: Before these changes this would have crashed if an env variable was missing (where the optional `dotenv.env[...]` was force unwrapped), now the value is set to an empty string (`String.fromEnvironment('I_DONT_EXIST') == ""`), i.e. we probably crash at some later point. Is this not a problem?

If it is we could use something like this

```dart
String? getEnv(String name) {
      return (dotenv.env[name] ?? (bool.hasEnvironment(name) ? String.fromEnvironment(name) : null))
}
```

And force-unwrap it to get the same behavior as before

jannis-baum (changes requested):

docker/docker-compose-app.yml
```diff
@@ -3,11 +3,11 @@ name: 'studyu'
 
 services:
   app:
-    container_name: app
+    container_name: studyu-app
```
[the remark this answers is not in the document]
johannesvedder: The compose name doesn't impact the container names afaik, so we do not get double studyu prefixes. The new prefix follows the way how supabase is naming their containers. With the changes the containers are now called `studyu-*`.
grafik

flutter_common/lib/src/utils/env_loader.dart
johannesvedder: In both ways the end result would probably be: "Crash, without any meaningful error message for the user, because this is still a TODO". But alright let's take the changes to keep the bad crash behavior consistently bad. :D

flutter_common/lib/src/utils/env_loader.dart
```diff
@@ -11,23 +11,30 @@ String envFilePath() {
...
     // optional
   );
   env.setEnv(
-    dotenv.env['STUDYU_SUPABASE_URL']!,
-    dotenv.env['STUDYU_SUPABASE_PUBLIC_ANON_KEY']!,
-    envAppUrl: dotenv.env['STUDYU_APP_URL'],
-    envDesignerUrl: dotenv.env['STUDYU_DESIGNER_URL'],
-    envProjectGeneratorUrl: dotenv.env['STUDYU_PROJECT_GENERATOR_URL'],
+    supabaseUrl,
+    supabaseAnonKey,
```
jannis-baum, line 34: Why is it possible to pass an optional String into `setEnv` here when it needs a non-optional String?👀

jannis-baum (approved):

flutter_common/lib/src/utils/env_loader.dart
johannesvedder: Dart runtime would already have crashed before if these weren't available:

```suggestion
    url: supabaseUrl!,
    anonKey: supabaseAnonKey!,
```

PR #479, opened 2023-07-07
Feature/image frontend in app

piipkufan:


johannesvedder:
Do you intend to merge this feature in its current state right now @piipkufan? I would suggest that we start reviewing and merging your group's whole project at once later in time, rather than every feature bit by bit.

PR #484, opened 2023-07-17
Issue/482 supabase test framework

johannesvedder:
- Fixes docker networks misbehaving
- Dumped study schema with postgres to verify it adheres to current version of supabase
  - Change table owners to postgres (fixes: https://github.com/orgs/supabase/discussions/9314)
- Provides the ability to test (supabase) postgresql functions and row level security with pgtap
- Requires the supabase CLI to be installed and can be initiated with `npx supabase test db`
- For manual testing,`database/studyu_schema.sql` has to be copied to `docker/supabase/seed.sql`. I didn't find a way to specify a different file for seeding
- depends on #483

Todo:

- [x] We could create a github actions pipeline for this if the current state is sufficient
  - https://2metz.fr/blog/use-pgtap-on-github-actions/

PR #485, opened 2023-07-17
Issue/483 bug study registry visibility issues

johannesvedder:
Changes can be tested with https://github.com/hpi-studyu/studyu/pull/484

Be careful when merging with #484 due to changed new studyu-schema.sql order.

### DIFF
--- a/core/lib/src/models/tables/study.dart
+++ b/core/lib/src/models/tables/study.dart
@@ -159,9 +159,8 @@ class Study extends SupabaseObjectFunctions<Study> {
   // ['id', 'title', 'description', 'published', 'icon_name', 'results', 'schedule']
   static Future<List<Study>> publishedPublicStudies() async {
     try {
-      return SupabaseQuery.extractSupabaseList<Study>(
-        await env.client.from(tableName).select(),
-      );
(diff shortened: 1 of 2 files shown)


core/lib/src/models/tables/study.dart
```diff
@@ -159,9 +159,8 @@ class Study extends SupabaseObjectFunctions<Study> {
   // ['id', 'title', 'description', 'published', 'icon_name', 'results', 'schedule']
   static Future<List<Study>> publishedPublicStudies() async {
     try {
-      return SupabaseQuery.extractSupabaseList<Study>(
-        await env.client.from(tableName).select(),
-      );
+      final response = await env.client.from(tableName).select().eq('participation', 'open') as List;
```
jannis-baum, line 162: Shouldn't this be checking for `registry_published` rather than `participation` now?
```suggestion
      final response = await env.client.from(tableName).select().eq('registry_published', true) as List;
```

johannesvedder: The method is used in the app. There we only want to show studies that can be accessed publicy i.e. participation=open.

jannis-baum (approved):

PR #490, opened 2023-07-27
Issue/489 refactor environments

johannesvedder:
- `.env.selfhost` is replaced with the custom environment `.env.hpi` to be used when deploying to studyu-hpi firebase
- `.env.staging` is replaced with `.env.dev` for deployment to studyu-dev firebase
- `.env.local` is now used for localhost development and basic selfhosting needs with docker compose


melos.yaml
jannis-baum, line 1: Some of these actions (e.g. `designer_v2:web:local`) have `designer_v2` in their name and others just `designer` (e.g. `dev:designer`) even though they're all for v2. I would suggest just naming all of them `designer` without `v2` since and removing the actions for v1 (or, in case they are still needed, suffixing them with `v1`)

I also think it'd be nice to follow some naming convention, e.g. `target:scope:whatever else we need` such as `dev:designer` (vs. `designer:web:local`)

melos.yaml
```diff
@@ -86,18 +86,18 @@ scripts:
   dev:designer:
     run: |
       melos exec -c 1 -- \
-        "flutter run -d chrome --web-port 3001 --dart-define=STUDYU_ENV=.env.selfhost"
+        "flutter run -d chrome --web-port 3001 --dart-define=STUDYU_ENV=.env.hpi"
     packageFilters:
       scope: studyu_designer_v2
-    description: Runs designer in Chrome on port 3001 with selfhost env for local development
+    description: Runs designer in Chrome on port 3001 with hpi env for local development
```
jannis-baum, line 92: I was using these workflows for development with my localhost Supabase instance before (i.e. to keep it the same, it should now be `.env.local`). Is `.env.hpi` the production environment? Is it safe to use this for local development?

jannis-baum (changes requested):

melos.yaml
johannesvedder: I'd recommend to use `.env.local` for now for local development as before. `.env` is currently the production environement and `.env.hpi` will be used for remote testing the current dev branch in the following weeks (but I might reset the db at any time so please do not store important stuff there). After migration, `.env.hpi` will replace the current `.env` environment, and `.env.dev` will be used for remote testing. That's the plan at least. :D

jannis-baum: Cool! Then let's switch the env for `dev:designer|app` to `.env.local`

melos.yaml
johannesvedder: We'll remove the legacy components in https://github.com/hpi-studyu/studyu/pull/495, so I'd move these suggestions there.

melos.yaml
```diff
@@ -86,18 +86,18 @@ scripts:
   dev:designer:
     run: |
       melos exec -c 1 -- \
-        "flutter run -d chrome --web-port 3001 --dart-define=STUDYU_ENV=.env.selfhost"
+        "flutter run -d chrome --web-port 3001 --dart-define=STUDYU_ENV=.env.local"
     packageFilters:
       scope: studyu_designer_v2
-    description: Runs designer in Chrome on port 3001 with selfhost env for local development
+    description: Runs designer in Chrome on port 3001 with hpi env for local development
```
jannis-baum, line 92:
```suggestion
    description: Runs designer in Chrome on port 3001 with local env for local development
```

Same below

jannis-baum (approved):

melos.yaml
johannesvedder: ups. fixed

jannis-baum (approved):

PR #492, opened 2023-07-29
Issue/491 feat studyu selfhost cli

johannesvedder:


docker/README.md
```diff
@@ -9,10 +9,52 @@ You will need to install Docker to be able to run any of the following scripts.
 Make sure you have [Docker](https://docker.com) installed and configured by
 running `docker -v`.
 
-For HPI VMs, [this
-script](https://gist.github.com/johannesvedder/0fafcbeabe8e069f96085bfedaebd9d0)
+For HPI VMs,
+[this script](https://gist.github.com/johannesvedder/0fafcbeabe8e069f96085bfedaebd9d0)
```
jannis-baum, line 13: Why not include this in this repo?

docker/README.md
```diff
@@ -9,10 +9,52 @@ You will need to install Docker to be able to run any of the following scripts.
...
+To get started quickly, run the StudyU CLI. It will interactively guide you
+through the setup process. The CLI is a wrapper around the docker-compose
+commands described in the following sections. It is located in the root
+directory of the repository and can be run with `./studyu-cli.sh`.
+
+During the configuration of the CLI, you will be asked to define the components
+you want to run. The following components are available:
+
+- `supabase-db`: A PostgreSQL database for Supabase
+- `supabase`: A self-hosted Supabase instance (needs `supabase-db`)
```
jannis-baum, line 30: Does it make sense to ever run these separately? If not, we could just combine them to simplify things

docker/README.md
```diff
@@ -9,10 +9,52 @@ You will need to install Docker to be able to run any of the following scripts.
...
+`studyu-designer`. If you only want to run a self-hosted Supabase
+instance and deploy StudyU elsewhere, you can choose `proxy` instead of
+`studyu-app` and `studyu-designer` to access Supabase Studio. Otherwise, if you
+only want to run the StudyU App and the StudyU Designer and not self-host Supabase,
+you can choose the respective component without `supabase-db` and `supabase`.
+Have a look into the use cases below for more details on the different
+configurations.
+
+```bash
+chmod +x studyu-cli.sh
```
jannis-baum, line 45: Why is this necessary? Can't we just commit the `+x` mode to the repo?

docker/README.md
```diff
@@ -9,10 +9,52 @@ You will need to install Docker to be able to run any of the following scripts.
...
+./studyu-cli.sh start
+```
+
+The CLI offers multiple commands to manage your StudyU instance. Run
+`./studyu-cli.sh --help` to see all available commands.
+
+After making changes to environment files such as `supabase/.env` or
+`flutter_common/lib/envs/.env.*`, a rebuild of the StudyU components might be
+necessary. You can run `./studyu-cli.sh delete` for this. Be sure to only remove
+the containers and images and not the volumes.
```
jannis-baum, line 56: How do we make sure to only remove the containers and images and not the volumes?

docker/studyu-cli.sh
```diff
@@ -0,0 +1,479 @@
+#!/bin/bash
```
jannis-baum, line 1: macOS sadly has a really old version of bash at `/bin/bash`. I would suggest using `/usr/bin/env bash` for the shebang which will use a newer version (if the user has one) and including a version check before running because otherwise people may get error messages they don't understand, such as

```plain
./studyu-cli.sh: line 17: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

for `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)`

docker/README.md
```diff
@@ -23,32 +65,46 @@ common use-cases to get you started quickly.
...
-If you want to use a different database, more steps are necessary. You can start
-a managed Supabase project at [supabase.com](https://supabase.com) and link it
-with your StudyU instance by replacing the values for `STUDYU_SUPABASE_URL` and
-`STUDYU_SUPABASE_PUBLIC_ANON_KEY` in the file `.env.local`. Alternatively, you
-can self-host Supabase on your own as explained in
+By default the main StudyU database will be used. If you want to use a different
+database, more steps are necessary. You can start a managed Supabase project at
+[supabase.com](https://supabase.com) and link it with your StudyU instance by
+replacing the values for `STUDYU_SUPABASE_URL` and
+`STUDYU_SUPABASE_PUBLIC_ANON_KEY` in the file `flutter_common/lib/envs/.env.local`.
```
jannis-baum, line 106: I think for this use-case it'd be best to recommend using some git-ignored file instead of this one because the person doing this will otherwise have changes to the repo

docker/README.md
```diff
@@ -100,17 +156,27 @@ Configuration](#advanced-configuration)**
 
 ### Run Supabase and StudyU
 
-Make sure that you thoughtfully completed all the previous steps and then start
-the following services:
+Make sure that you thoughtfully completed all the previous steps. You can either
+use the StudyU CLI or manually run the docker-compose commands.
+
+StudyU CLI:
```
jannis-baum, line 162: My linter would say that this is a level 4 heading🙊😜 same as the ones below

docker/studyu-cli.sh
```diff
@@ -0,0 +1,479 @@
...
+available_components+="  proxy              Nginx reverse proxy to access the Supabase frontend if no StudyU component is started\n"
+# available_components+="  studyu             StudyU App and StudyU Designer (includes proxy)\n"
+available_components+="  studyu-app         StudyU App (includes proxy)\n"
+available_components+="  studyu-designer    StudyU Designer (includes proxy)"
+
+# Function to read configuration values from the file
+read_config() {
+  local failOnError="${1:-true}"
+
+  if [ "$failOnError" = true ] && [ ! -f "$CONFIG_FILE" ]; then
```
jannis-baum, line 37: the command `[` has an `-a` flag for logical `and`s

```suggestion
  if [ "$failOnError" = true -a ! -f "$CONFIG_FILE" ]; then
```

or, better, use bash's `[[`

```suggestion
  if [[ "$failOnError" = true && ! -f "$CONFIG_FILE" ]]; then
```

Same for all the other conditions; when we don't need POSIX compliance (e.g. writing a bash or zsh script) it's usually easier to work with `[[`

docker/studyu-cli.sh
```diff
@@ -0,0 +1,479 @@
...
+  fi
+
+  cd "$STUDYU_PATH/docker/" || exit 1
+}
+
+
+# Print CLI configuration file
+print_config() {
+  local failOnError="${1:-true}"
+  read_config failOnError
```
jannis-baum, line 62: Probably meant to be

```suggestion
  read_config $failOnError
```

docker/studyu-cli.sh
jannis-baum, line 1: The first time running this as `./studyu-cli.sh config` gives me

```
studyu-cli.sh: line 46: <my path>/studyu/docker/.studyu_config: No such file or directory
studyu-cli.sh: line 56: cd: /docker/: No such file or directory
```

Because it (1) tries to read the config file which doesn't yet exist and then (2) tries to cd into `"$STUDYU_PATH/docker/"` which is just `/docker/` because `$STUDYU_PATH` is not set since there is no config file that would have been used to set it :(

---

If I set `STUDYU_PATH` up correctly manually and then run it, the config works but I get stuck trying to `start`:

```
Failed to load <my path>/studyu/docker/flutter_common/lib/envs/.env.local:
  open <my path>/studyu/docker/flutter_common/lib/envs/.env.local: no such file or directory
```

Looks like there is another issue with the relative/absolute paths. I can keep testing when these issues are fixed :)

docker/studyu-cli.sh
```diff
@@ -0,0 +1,479 @@
...
+  set +a
+
+  # Check if the required variables are present
+  if [ "$failOnError" = true ] && { [ -z "$STUDYU_DEFAULT_COMPONENTS" ] || [ -z "$STUDYU_PATH" ] || [ -z "$STUDYU_BASE_BRANCH" ]; }; then
+    echo -e "${red}Error: Configuration file $CONFIG_FILE is missing some entries.${reset}"
+    echo "Please run 'config' to set the required options."
+    exit 1
+  fi
+
+  cd "$STUDYU_PATH/docker/" || exit 1
```
jannis-baum, line 55: Oh and another thing: why is there a need to set STUDYU_PATH manually at all? The script's parent dir is always correct, isn't it? Moving the script would break many things anyways.

docker/README.md
johannesvedder: Thought of this as well, but splitting it up has the advantages that Supabase can be hosted and maintained independently from the db, which brings further benefits regarding upgrading, load-balancing, etc.

docker/README.md
johannesvedder: Don't know what the best-practise is here, but if there are no downsides, absolutely.

jannis-baum: I always do it like this and haven't come across any downsides thus far🙈

docker/studyu-cli.sh
johannesvedder: 👍 Do you think something like this would be sufficient?

```
version_above_4(){
    [ -z $BASH_VERSION ] && return 1
    case $BASH_VERSION in 
        4.*) return 0 ;;
        ?) return 1;; 
    esac
}
if  ! version_above_4; then
    echo "Bash version equal or larger than 4 is required to run this script."
fi
```

docker/README.md
johannesvedder: They'll have changes in the repo with the current setup as well when modifying ports in the docker compose scripts, customizing domains in the nginx config files, or switching the the login data in the .htpasswd file. This is far from optimal at the moment. Maybe those changes can be outsourced to the `docker/.studyu_config` file and injected with the cli at startup. But these are future todos. Having a repo with no changes is out of scope atm.

docker/README.md
johannesvedder: I tried that but then the linter said no equal headings in one file haha. I can enumerate the headings if you think that'd be a better idea. Nevertheless, we should move the documentation [to the website](https://www.studyu.health/docs/development/selfhosting) at some point to split it up and increase readability

docker/studyu-cli.sh
johannesvedder: I think I changed them all 👍

docker/studyu-cli.sh
johannesvedder: I thought of adding the possibility in the future to install the script to a PATH dir in order to make it run globally as a `studyu` command. After setting the STUDYU_PATH this might be possible.

docker/studyu-cli.sh
johannesvedder: If there's no config file the file shouldn't be read anymore when running `config`. After configuration `STUDYU_PATH` is present now.

docker/README.md
johannesvedder: We probably don't want to make recommendations on how to install docker for the general majority of people running this repo outside of HPI perimeter's.

docker/studyu-cli.sh
jannis-baum: Like this it would fail for versions > 4, is that intended? I only have that old bash version and `5.2.15(1)-release` on my machine. Anything starting at version 4 is okay, I would suggest

```bash
if [[ "$(head -c 1 <<< $BASH_VERSION)" -lt 4 ]]; then
    echo "Bash version equal or larger than 4 is required to run this script."
    exit 1
fi
```

This also still covers the case that bash is not the shell running the script since `"" < 4` :)

docker/README.md
jannis-baum: Oki I'll add this to #499

docker/README.md
jannis-baum: Oh true! I'd be good with turning off the rule for no equal headings, I don't like that rule. You can decide if you want to turn it off or leave it as is :)

docker/studyu-cli.sh
jannis-baum: Okay, that makes sense. But then we also (1) need to move the config to something like `~/.studyu_config` or (2) rely on `STUDYU_PATH` already being set ahead of time.

docker/studyu-cli.sh
jannis-baum, line 1: A though on the name of this file: `.sh` makes it look like this is an `sh` script, which might make someone try to run `sh studyu-cli.sh`. As such, I would rename it to `.bash` for being a `bash` script.

However, since this file is an executable with a shebang I think it's best to just give it no file extension at all, i.e. just `studyu-cli`. Being executable and having a shebang & version check, no one should have to know what interpreter needs to be used to run this.

For additional reading, [this person on stackoverflow](https://stackoverflow.com/a/23167175) shares my opinion.

docker/studyu-cli.sh
johannesvedder: Since the script should be usable by multiple users on the same machine, relying on home directories is no good. However, we can rely on `STUDYU_PATH` being set, since it is part of the configuration file `.studyu_config` that is generated at first launch.

```
❯ cat .studyu_config
STUDYU_DEFAULT_COMPONENTS="supabase-db supabase studyu-app studyu-designer"
STUDYU_PATH="/home/Johannes/studyu"
STUDYU_BASE_BRANCH="dev"
```

Without `STUDYU_PATH` being set, no action can be performed, since this check inside `read_config` is executed before running any command:

https://github.com/hpi-studyu/studyu/blob/3692ccdb767580eb6f579730f5e92f59023b93b4/docker/studyu-cli.sh#L49-L53

When running `config` the first time, the configuration file will be created at (a more or less hardcoded) path.

https://github.com/hpi-studyu/studyu/blob/3692ccdb767580eb6f579730f5e92f59023b93b4/docker/studyu-cli.sh#L14

If someone would like to move `studyu-cli.sh` to a directory that is not writable by the user, setting the `STUDYU_CONFIG_FILE` env variable and recreating the config or move the old one to the new position should be sufficient. I added this:

```
# Configuration file path
# Check if the STUDYU_CONFIG_FILE environment variable is set
if [ -n "$STUDYU_CONFIG_FILE" ]; then
    CONFIG_FILE="$STUDYU_CONFIG_FILE"
else
    # Otherwise, fallback to the hardcoded path
    SCRIPT_DIR="$(dirname "$0")"
    CONFIG_FILE="$SCRIPT_DIR/.studyu_config"
fi
```

jannis-baum (dismissed):
I committed the `+x` mode, hope that's okay!

See one final comment & then we can merge🥳

---

Some other last thoughts: It might be useful to add

1. something like `studyu-cli status` which exits with `0`/`1` depending on whether the containers are running & okay, kind of like a health check
2. ~~this might be a bit out there but I think it'd be super cool for development and anyone using this with an interactive shell: Modifying the `$PROMPT` when starting/stopping the containers with the CLI (like Python's venv activate/deactivate for example) so that it's clear that we're still running StudyU and it's harder to forget to stop the detached containers.~~ EDIT: forgot that we're not sourcing but executing the script; this is too much work for the value. If we want this it'd be easier to have people add it to their `zsh/bashrc` themselves using `studyu-cli status` from 1. For that we could add example code somewhere in the docs

If these are out of scope for this PR, let's open an issue/issues for them.

jannis-baum:
We can also remove `.meta/kitty-dev` in this PR since the CLI makes it redundant

docker/studyu-cli
```diff
@@ -0,0 +1,516 @@
...
+  for name in $container_names; do
+      health_status=$(docker inspect --format='{{.State.Running}}' "$name")
+      if [[ "$health_status" != "true" ]]; then
+          echo -e "${red}Error: $name is not running healthy. Check 'docker compose ps' for troubleshooting.${reset}"
+          exit 1;
+      fi
+  done
+  echo -e "${green}All components are running healthy.${reset}"
+  exit 0;
+}
```
jannis-baum, line 299: This says everything is running healthy when no containers are running; maybe we can first verify that everything that is configured to run is running at all

PR #495, opened 2023-07-30
release(2023-07-31): Release preparations

johannesvedder:
- Upgrade deps
- Remove legacy components (designer, repo_generator, notebook_uploader)
- Make analytics fully optional and migratable

---

## After this PR no more features will be incorporated into the next update


database/migrate_analytics.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.app_config
+ADD COLUMN analytics jsonb;
```
jannis-baum, line 2: newline is missing at end of file

jannis-baum (commented):
nice🧹

Now that version 1 of the designer is gone, can't we rename `designer_v2` to just `designer`?

johannesvedder:
> nice🧹
> 
> Now that version 1 of the designer is gone, can't we rename `designer_v2` to just `designer`?

Yep, but I'd prefer doing that in a separate step after the next update.

jannis-baum (approved):

PR #514, opened 2023-10-21
Fix #513 (navigation bug)

### LINKED ISSUE #513: Black screen after tapping "Back" in Study Selection
See https://trello.com/c/KUD6TOj6/

hig-dev:
The previous code:

`Navigator.pushNamedAndRemoveUntil(context, Routes.studySelection, (_) => false)`

removes the dashboard from the navigation stack.

After tapping "Back" in the study selection page, the navigation stack is empty resulting in the black screen.
I also checked if it was necessary to remove pages from the navigation when the dashboard is presented after ending a study but that was not the case.


johannesvedder (approved):
I tested the old code, verified the bug, and ran a new study with the fixed PR. Going back from study selection works. Feel free to merge. Nice catch! 🐟

PR #530, opened 2023-11-16
fix: add invoker to study_progress_export view

johannesvedder:


hig-dev (dismissed):
Quick approval as it is a hotfix for a security issue.
We should add tests to prevent this in the future.

hig-dev (approved):

PR #532, opened 2023-11-17
Fix/app running parallel

johannesvedder:


johannesvedder:
Might get fixed temporarily with https://github.com/hpi-studyu/studyu/pull/545

PR #538, opened 2023-11-29
chore: Do not allow the password_recovery page to be publicly accessible

johannesvedder:


designer_v2/lib/routing/router_config.dart
```diff
@@ -359,6 +349,16 @@ class RouterConf {
             const AccountSettingsDialog(),
           );
         }),
+    GoRoute(
```
hig-dev, line 352: I understand this change but I don't understand why the private routes are private. The only usage that I could find is here:

https://github.com/hpi-studyu/studyu/blob/c44aef505115e6cc2718fca048bc3171d4b694d2/designer_v2/lib/routing/router_config.dart#L65

It looks like there is no difference between private and public routes.

PR #554, opened 2023-12-20
Git: Normalize line endings to LF for text files

hig-dev:
This fixes the issue with generated files that have the same content but different line endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,14 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c text
+*.h text
+
+# Declare files that will always have CRLF line endings on checkout.
+*.sln text eol=crlf
+
+# Denote all files that are truly binary and should not be modified.
+*.png binary
+*.jpg binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,11 @@
 # to native line endings on checkout.
 *.c text
 *.h text
+*.dart text
+*.sh text
+*.sql text
+*.md text
+*.svg text -diff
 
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
johannesvedder, line 1: Let's follow GitHub's [recommendations](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=linux#example) just to be on the safe side.

```suggestion
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# Explicitly declare text files you want to always be normalized and converted
# to native line endings on checkout.
*.c text
*.h text

# Declare files that will always have CRLF line endings on checkout.
*.sln text eol=crlf

# Denote all files that are truly binary and should not be modified.
*.png binary
*.jpg binary

```

And having a newline at the EOF is also good to be POSIX compliant.

johannesvedder (changes requested):

.gitattributes
```diff
@@ -0,0 +1,14 @@
...
+# to native line endings on checkout.
+*.c text
+*.h text
+
+# Declare files that will always have CRLF line endings on checkout.
+*.sln text eol=crlf
+
+# Denote all files that are truly binary and should not be modified.
+*.png binary
+*.jpg binary
```
johannesvedder, line 14: Just one more thing: Ignore generated documentation files and other svg files in diffs

```suggestion
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# Explicitly declare text files you want to always be normalized and converted
# to native line endings on checkout.
*.c text
*.h text
*.dart text
*.sh text
*.sql text
*.md text
*.svg text -diff

# Declare files that will always have CRLF line endings on checkout.
*.sln text eol=crlf

# Denote all files that are truly binary and should not be modified.
*.png binary
*.jpg binary
*.svg binary

# Files that should be ignored for the repository's language statistics and hidden by default in diffs.
docs/database/** linguist-generated=true
docs/uml/** linguist-generated=true
```

johannesvedder (approved):

Issue #517, opened 2023-11-01
Bug: Sorting is broken on study registry


hig-dev:
See https://trello.com/c/9sQkpxXH/101-bug-sorting-is-broken-on-study-registry

Issue #516, opened 2023-11-01
Implement studies table with ListView


hig-dev:
See https://trello.com/c/AKEq97B1/100-implement-studies-table-with-listview

Issue #497, opened 2023-07-30
Bug: Designer breaks when resizing window


johannesvedder:
See [Trello card](https://trello.com/c/yxAe26eh)

jannis-baum:
Was closed in #498

hig-dev:
Still happens for study registry

PR #555, opened 2023-12-21
Feat/studiestablelistview: Implement studies table with ListView

### LINKED ISSUE #556: Dashboard page looks bad with small horizontal space
I think we also need to add a horizontal scrollbar:

image

_Originally posted by @hig-dev in https://github.com/hpi-studyu/studyu/issues/518#issuecomment-1811730928_

hig-dev:
The current implementation uses a Table which is not suitable for future additions such as expandable rows.

The new implementation also fixes resizing bugs and sorting.

Fixes https://github.com/hpi-studyu/studyu/issues/516
Fixes https://github.com/hpi-studyu/studyu/issues/497 (for study registry)
Fixes https://github.com/hpi-studyu/studyu/issues/517

In addition, the columns Entrolled, Active, and Completed are collapsed if the horizontal space is too low.

Fixes #556


hig-dev:
Issue #556 should be fixed with this PR #555 and PR #557

johannesvedder:
Does this depend on #557?

johannesvedder (commented):
Nice PR for switching to a ListView implementation. We should try to fix the horizontal width issues before merging this. The menu widgets seem to overflow a lot, especially in the German version. My additions left [here](https://github.com/hpi-studyu/studyu/pull/557#pullrequestreview-1794372425) do also occur. Are both PR's depending on each other?

grafik

hig-dev:
> Does this depend on #557?

#557 definitively helps with the overflowing issues. So I would suggest to merge it first.

hig-dev:
I merged #557 into this PR and fixed the overflowing issues.

johannesvedder:
> I merged #557 into this PR and fixed the overflowing issues.

Can we close #557 then?

johannesvedder:
And what do you think about the other issues I posted [here](https://github.com/hpi-studyu/studyu/pull/557#pullrequestreview-1794372425)?

hig-dev:
> And what do you think about the other issues I posted [here](https://github.com/hpi-studyu/studyu/pull/557#pullrequestreview-1794372425)?

> Small QoL improvement: Close the drawer in compact mode after a selection has been made

Yes, that's done.

> The available space for the "Created" column becomes smaller and smaller as the page width decreases until the column has disappeared completely.
> Similar behavior with the most right columns as they are just cut off with decreasing screen widths. This does not only impair visual appearance, but also has an impact on usability, as the menu can not be used on smaller screen widths with non-collapsed drawer. To counteract this, we would either need to change the compact threshold to a very high value, or add a horizontal scrollbar. I would recommend the latter.

I added one additional threshold to collapse the created column, and increased the other threshold. So these issues should be resolved. I really would like to avoid a horizontal scrollbar.

johannesvedder:
Horizontal scaling feels great now! 👍 

Some issues I noticed comparing the old view vs. the new one.

## Old Table
image

## New Table


image

- "New study" button got smaller. Was that intentional? If not, should we increase the size again or leave it like this?
- The alignment of the sorting arrows is faulty. The right alignment leads to different positions related to the title element. E.g. for status, participation and created it is unclear which arrow belongs to which item. Perhaps use a fixed margin instead, similar to the old table.
image
- The default sorting option is not available anymore. The arrows can only be toggled up or down. Having a state where all arrows are disabled (compare to Spotify playlists) was intentional in order to implement a custom sorting feature on top of it.
- The design of the list items looks intrusive to me. The top and bottom border both have different styles which is not used in the Designer elsewhere. I'd suggest switching back to the previous design to keep the dashboard flat and streamlined.

hig-dev:
> "New study" button got smaller. Was that intentional? If not, should we increase the size again or leave it like this?

It was not intentional, I increased the height so that it matches with the previous version: https://github.com/hpi-studyu/studyu/pull/555/commits/58c731aea4352956446ab884e52c18d2f6cc0715

> The alignment of the sorting arrows is faulty. The right alignment leads to different positions related to the title element. E.g. for status, participation and created it is unclear which arrow belongs to which item. Perhaps use a fixed margin instead, similar to the old table.

Nice catch, fixed in https://github.com/hpi-studyu/studyu/pull/555/commits/1f709f54cf29d7630802d20901c5ffe3b4e16156

> The default sorting option is not available anymore. The arrows can only be toggled up or down. Having a state where all arrows are disabled (compare to Spotify playlists) was intentional in order to implement a custom sorting feature on top of it.

As long as there is no custom sorting feature, I would suggest to always have a sorting. The default sorting could be the title or created time. See also: https://github.com/hpi-studyu/studyu/pull/457#pullrequestreview-1578326828

> The design of the list items looks intrusive to me. The top and bottom border both have different styles which is not used in the Designer elsewhere. I'd suggest switching back to the previous design to keep the dashboard flat and streamlined.

I removed the elevation from the study table items (https://github.com/hpi-studyu/studyu/pull/555/commits/2bd24a6b98c35682059525b914c869385dfdd21d) for a flatter design.  The new design is still different from the old one but I think it is an improvement. Feel free to share your opinion or suggestions.

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -31,142 +67,168 @@ class StudiesTable extends StatelessWidget {
...
-      }
-      return 0;
-    };
-  }
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        final isCompact = constraints.maxWidth < compactWidthThreshold;
+        final isSuperCompact = constraints.maxWidth < superCompactWidthThreshold;
+        final isCompactStatTitle = constraints.maxWidth < compactStatTitleThreshold;
+        debugPrint('studies table width: ${constraints.maxWidth}');
```
johannesvedder, line 79:
```suggestion
```

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -31,142 +67,168 @@ class StudiesTable extends StatelessWidget {
...
-      (Study a, Study b) => a.participation.index.compareTo(b.participation.index),
-      (Study a, Study b) => a.createdAt!.compareTo(b.createdAt!),
-      (Study a, Study b) => a.participantCount.compareTo(b.participantCount),
-      (Study a, Study b) => a.activeSubjectCount.compareTo(b.activeSubjectCount),
-      (Study a, Study b) => a.endedCount.compareTo(b.endedCount),
-    ];
-    return predicates;
-  }
+        // Calculate the minimum status column width
+        int maxStatusLength = "Entwurf".length;
```
johannesvedder, line 86: Should this be localized?

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -31,142 +67,168 @@ class StudiesTable extends StatelessWidget {
...
+        maxStatusLength = max(maxStatusLength, tr.studies_list_header_status.length);
+        final double statusColumnWidth = maxStatusLength * 11.5;
 
-  List<Widget> _buildRow(BuildContext context, Study item, int rowIdx, Set<MaterialState> states) {
-    final theme = Theme.of(context);
+        // Calculate the minimum participation column width
+        final int maxParticipationLength = isCompact ? "Invite-only".length : tr.participation_invite_who.length;
```
johannesvedder, line 91: Should this be localized?

johannesvedder (approved):

designer_v2/lib/features/dashboard/studies_table.dart
hig-dev: No, it is a design choice what this lenght should be. The string just gives a hint why the specific length was chosen.

PR #557, opened 2023-12-21
Feat/collapsedashboardmenu: Make dashboard page responsive

### LINKED ISSUE #556: Dashboard page looks bad with small horizontal space

hig-dev:
- Collapses dashboard drawer if horizontal space is too low
- Make width of search bar responsive

Fixes #556


designer_v2/lib/features/dashboard/dashboard_scaffold.dart
```diff
@@ -5,17 +5,25 @@ import 'package:studyu_designer_v2/features/app_drawer.dart';
...
+    final isCompact = size.width < compactWidthThreshold;
     return Scaffold(
         backgroundColor: Theme.of(context).scaffoldBackgroundColor.faded(0.75),
-        appBar: null, // default app bar not suitable for our layout
+        appBar: isCompact
+            ? AppBar(
+                title: Text(tr.studyu),
+              )
+            : null,
+        drawer: isCompact ? AppDrawer(title: tr.studyu) : null,
```
johannesvedder, line 24: Using the logo might be more appropriate here instead of showing the (unthemed) text.

johannesvedder (changes requested):
Great addition for smaller screen widths! I like it. There are just some issues that I noticed:
- Small QoL improvement: Close the drawer in compact mode after a selection has been made
- The available space for the "Created" column becomes smaller and smaller as the page width decreases until the column has disappeared completely.
- Similar behavior with the most right columns as they are just cut off with decreasing screen widths. This does not only impair visual appearance, but also has an impact on usability, as the menu can not be used on smaller screen widths with non-collapsed drawer. To counteract this, we would either need to change the compact threshold to a very high value, or add a horizontal scrollbar. I would recommend the latter.

designer_v2/lib/features/dashboard/dashboard_scaffold.dart
johannesvedder: See: https://github.com/hpi-studyu/studyu/pull/560

designer_v2/lib/features/app_drawer.dart
```diff
@@ -81,55 +90,67 @@ class AppDrawer extends ConsumerStatefulWidget {
...
+          onSelected: (context, ref) {
+            ref.read(authRepositoryProvider).signOut();
+          },
+        ),
+      ],
+    ];
+  }
+
   @override
   void didUpdateWidget(AppDrawer oldWidget) {
```
hig-dev, line 155: Commit b01544bd breaks the update of the selected route.
That's why I reverted it. 

@johannesvedder : Do you have a better idea for the update of the selected route that works with a `const AppDrawer`?

hig-dev:
> * Small QoL improvement: Close the drawer in compact mode after a selection has been made

Done.

hig-dev:
Will be merged with #555

PR #562, opened 2024-01-17
fix: type checks

hig-dev:
Using `runtimeType` for type checks ist not recommended and I experienced a bug related to that.
This PR uses the pattern matching feature from Dart 3.0.

More info on type matching is here: https://github.com/dart-lang/language/issues/2911#issuecomment-1468272431


app/lib/screens/study/tasks/task_screen.dart
```diff
@@ -34,23 +34,22 @@ class _TaskScreenState extends State<TaskScreen> {
...
         return QuestionnaireTaskWidget(
-          task: taskInstance.task as QuestionnaireTask,
+          task: questionnaireTask,
           key: UniqueKey(),
           completionPeriod: taskInstance.completionPeriod,
         );
       default:
-        print('${taskInstance.task.runtimeType} is not a supported Task!');
-        return null;
+        throw ArgumentError('Task ${taskInstance.task.type} not supported');
```
hig-dev, line 52: This default switch branch can be prevented in the future if we use sealed classes: https://dart.dev/language/class-modifiers#sealed

app/lib/widgets/questionnaire/question_container.dart
```diff
@@ -26,46 +26,45 @@ class _QuestionContainerState extends State<QuestionContainer> with AutomaticKee
...
+      case AnnotatedScaleQuestion annotatedScaleQuestion:
         return AnnotatedScaleQuestionWidget(
-          question: widget.question as AnnotatedScaleQuestion,
+          question: annotatedScaleQuestion,
           onDone: _onDone,
         );
       default:
-        print('Question not supported!');
-        return null;
+        throw ArgumentError('Question type ${widget.question.type} not supported');
```
hig-dev, line 60: This default switch branch can be prevented in the future if we use sealed classes: https://dart.dev/language/class-modifiers#sealed

app/lib/screens/study/tasks/task_screen.dart
johannesvedder: https://github.com/hpi-studyu/studyu/issues/566

app/lib/widgets/questionnaire/question_container.dart
johannesvedder: https://github.com/hpi-studyu/studyu/issues/566

johannesvedder (approved):
Nice 👍

PR #563, opened 2024-01-17
fix: foreground color of complete button

hig-dev:
The complete button looked like this:
image

So I changed the foreground color to white.


johannesvedder:
I have already fixed this together with other design issues in a separate branch. I'll open a PR soon. 👍

PR #571, opened 2024-01-26
feat: add ready_to_merge.yml workflow

johannesvedder:
This PR introduces a github action that checks if all required checks have passed before merging is allowed.
A workaround with a separate action is necessary, since Github's branch protection requires all checks to pass even if they haven't been run before.
For more information see: https://github.com/orgs/community/discussions/13690

Additionally, some pipeline related bugs are fixed.


hig-dev (approved):
That is a nice addition.

PR #576, opened 2024-02-05
feat: check min app version and force update

hig-dev:
This PR allows to set a minimum app version in the database.

Hint: the app only checks for major and minor version, and not the patch version.

Example: 2.1.12 (major: 2, minor: 1, patch: 12 <- ignored)


database/studyu-schema.sql
```diff
@@ -491,6 +491,9 @@ ALTER FUNCTION public.user_email(user_id uuid) OWNER TO postgres;
 
 CREATE TABLE public.app_config (
     id text NOT NULL,
+    app_min_version text NOT NULL DEFAULT '0.0.0',
```
johannesvedder, line 494: I think we should not set a default here, but require this field to be set on table creation. The recommended "default value" would then be the current app version or the lowest version without a breaking database change. We will need to keep track of documenting app_min_version somewhere when we release.

app/lib/l10n/app_en.arb
```diff
@@ -188,5 +188,7 @@
       "placeholders": {
       "pattern": {}
       }
-  }
+  },
+  "app_outdated_message": "Your app is outdated.\nPlease update to the latest version.",
```
johannesvedder, line 192: A new version of the StudyU app is available. Please update it to receive the latest features and improvements. Thank you for your support!

app/lib/l10n/app_de.arb
```diff
@@ -189,5 +189,7 @@
     "placeholders": {
       "pattern": {}
     }
-  }
+  },
+  "app_outdated_message": "Ihre App ist veraltet.\nBitte aktualisieren Sie auf die neueste Version.",
```
johannesvedder, line 193: Eine neue Version der StudyU App ist verfügbar. Bitte führen Sie eine Aktualisierung durch, um die neuesten Funktionen und Verbesserungen zu erhalten. Vielen Dank für Ihre Unterstützung!

database/migration/migrate-app_min_version.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE public.app_config
+ADD COLUMN app_min_version text NOT NULL DEFAULT '0.0.0';
+
+ALTER TABLE public.app_config
+ADD COLUMN app_playstore_url text NOT NULL DEFAULT 'https://play.google.com/store/apps/details?id=health.studyu.app';
+
+ALTER TABLE public.app_config
+ADD COLUMN app_appstore_url text NOT NULL DEFAULT 'https://itunes.apple.com/app/id1571991198';
```
johannesvedder, line 8:
```suggestion
```

database/migration/migrate-app_min_version.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE public.app_config
+ADD COLUMN app_min_version text NOT NULL DEFAULT '0.0.0';
```
johannesvedder, line 2:
```suggestion
ADD COLUMN app_min_version text NOT NULL;
```

app/lib/screens/app_onboarding/app_outdated_screen.dart
```diff
@@ -0,0 +1,61 @@
...
+              ),
+              const Spacer(),
+              storeUrl != null && storeIcon != null
+                  ? OutlinedButton.icon(
+                      icon: Icon(storeIcon),
+                      onPressed: () async {
+                        await launchUrl(Uri.parse(storeUrl!),
+                            mode: Platform.isAndroid
+                                ? LaunchMode.externalNonBrowserApplication
+                                : LaunchMode.platformDefault);
```
johannesvedder, line 49: Why not using `LaunchMode.externalNonBrowserApplication` for all platforms?

johannesvedder (changes requested):

app/lib/main.dart
```diff
@@ -36,20 +38,41 @@ Future<void> main() async {
...
   usePathUrlStrategy();
   AppConfig? appConfig;
+  String initialRoute = Routes.loading;
   try {
     appConfig = await AppConfig.getAppConfig();
+    if (await isAppOutdated(appConfig)) {
+      initialRoute = Routes.appOutdated;
+    }
   } catch (error) {
     // device could be offline
```
johannesvedder, line 48: Add a debug output for analysis. Might also be worth to track the error case to catch malformed `app_min_version` strings.

database/studyu-schema.sql
```diff
@@ -491,6 +491,9 @@ ALTER FUNCTION public.user_email(user_id uuid) OWNER TO postgres;
...
+    app_playstore_url text NOT NULL DEFAULT 'https://play.google.com/store/apps/details?id=health.studyu.app',
+    app_appstore_url text NOT NULL DEFAULT 'https://itunes.apple.com/app/id1571991198',
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

database/studyu-schema.sql
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

app/lib/l10n/app_en.arb
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

database/migration/migrate-app_min_version.sql
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

database/migration/migrate-app_min_version.sql
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

app/lib/main.dart
hig-dev: I moved the `isAppOutdated` call outside of the try catch block.

Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/8d36dc78771d748483bbf3ad22794a95e73dad06

app/lib/screens/app_onboarding/app_outdated_screen.dart
hig-dev: It was like that in https://github.com/larryaasen/upgrader/blob/212c23248ef9a2f8c02bce0daa5459cc7ecd8cf6/lib/src/upgrader.dart#L633 but I am not sure if that is neccessary.

I changed it to only use `LaunchMode.externalNonBrowserApplication` but I did not test it with a physical iOS device.

database/migration/migrate-app_min_version.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.app_config
+ADD COLUMN app_min_version text NOT NULL;
```
johannesvedder, line 2: Let's stick with newlines at the EOF for POSIX compliance, in case the sql file gets piped into a script.

app/lib/screens/app_onboarding/app_outdated_screen.dart
```diff
@@ -0,0 +1,57 @@
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import 'package:material_design_icons_flutter/material_design_icons_flutter.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+class AppOutdatedScreen extends StatelessWidget {
+  static const String kPlayStoreUrl = 'https://play.google.com/store/apps/details?id=health.studyu.app';
+  static const String kAppstoreUrl = 'https://itunes.apple.com/app/id1571991198';
```
johannesvedder, line 10: Small thing, but for the sake of consistency let's start a `constants.dart` similar as the one for the Designer for the App.

johannesvedder (commented):
Code looks fine, I'll test it in the next days and then we can merge :)

johannesvedder:
What happens if the app is not closed but remains open in the background? Do we need to address this case?

hig-dev:
> What happens if the app is not closed but remains open in the background? Do we need to address this case?

I don't see a problem with that.

johannesvedder:
- [ ] Todo fix https://github.com/hpi-studyu/studyu/actions/runs/7817193502/job/21324494867

app/lib/screens/app_onboarding/app_outdated_screen.dart
```diff
@@ -0,0 +1,55 @@
...
+    String? storeUrl;
+    IconData? storeIcon;
+    if (Platform.isAndroid) {
+      storeUrl = playStoreUrl;
+      storeIcon = MdiIcons.googlePlay;
+    } else if (Platform.isIOS) {
+      storeUrl = appstoreUrl;
+      storeIcon = MdiIcons.apple;
+    }
+    return Scaffold(
```
johannesvedder, line 24: Include an else to handle the web instance and other platforms we might support in the future. In that case do not show any url.

app/lib/screens/app_onboarding/app_outdated_screen.dart
```diff
@@ -0,0 +1,55 @@
...
+      body: SafeArea(
+        child: Center(
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: <Widget>[
+              const Spacer(),
+              const Image(image: AssetImage('assets/images/icon_wide.png'), height: 200),
```
johannesvedder, line 31:
```suggestion
              const Image(image: AssetImage('assets/icons/logo.png'), height: 200),
```

app/lib/l10n/app_en.arb
```diff
@@ -188,5 +188,7 @@
       "placeholders": {
       "pattern": {}
       }
-  }
+  },
+  "app_outdated_message": "A new version of the StudyU app is available. Please update it to receive the latest features and improvements. Thank you for your support!",
+  "open_store": "Open Store"
```
johannesvedder, line 193: If we don't want to refer to the actual store names to ease localization, let's phrase this differently since "Open Store" could be perceived as a specific term. Better use something like "Update now".

app/lib/screens/app_onboarding/app_outdated_screen.dart
johannesvedder: `LaunchMode.externalNonBrowserApplication` does not work using my iOS simulator. Nothing happens. Maybe let's use `LaunchMode.externalApplication` as described [here](https://stackoverflow.com/a/72532633). This does at least open the link on my simulator. I'll need to check with a physical device.

johannesvedder (changes requested):

johannesvedder:
> > What happens if the app is not closed but remains open in the background? Do we need to address this case?
> 
> I don't see a problem with that.

We only handle the version check during the first load of the app. So for users who have the app already open at the time the app_min_version is updated, they will not get notified about the update until the next restart of the app and likely run into issues. Another edge case would be users who rely on the web version of the app and use a bookmark for a static page of the app (e.g. app.studyu.health redirects to the /welcome page). When accessing static pages directly, the version check will never be executed.

app/lib/screens/app_onboarding/app_outdated_screen.dart
hig-dev: The url will not be shown if the platform is not Android or iOS. So no change is needed.

hig-dev:
With https://github.com/hpi-studyu/studyu/pull/576/commits/e82c3ac68beb4adf0662753c551d7cf3d53814b9 the app rechecks the min version on resume of the app. This should also work with bookmarks.

app/lib/screens/app_onboarding/app_outdated_screen.dart
johannesvedder: Other platforms will currently throw an exception. We should handle that:

```
======== Exception caught by widgets library =======================================================
The following UnsupportedError was thrown building AppOutdatedScreen(dirty, dependencies: [_LocalizationsScope-[GlobalKey#40f19]]):
Unsupported operation: Platform._operatingSystem

The relevant error-causing widget was: 
  AppOutdatedScreen AppOutdatedScreen:studyu/app/lib/routes.dart:66:56
When the exception was thrown, this was the stack: 
[...]
==> packages/studyu_app/screens/app_onboarding/app_outdated_screen.dart 17:18         build
```

johannesvedder (changes requested):
https://github.com/hpi-studyu/studyu/pull/576#discussion_r1514487863 and https://github.com/hpi-studyu/studyu/pull/576#discussion_r1508951713 will need changes, looks good otherwise.

app/lib/screens/app_onboarding/app_outdated_screen.dart
hig-dev: @johannesvedder Can you check with a physical device?

app/lib/screens/app_onboarding/app_outdated_screen.dart
hig-dev: Fixed in https://github.com/hpi-studyu/studyu/pull/576/commits/40f39afe60293bfd62ea705a55f59012c6aa00e3

johannesvedder (approved):

PR #583, opened 2024-02-06
Issue/564 future proof deserialization with exceptions v2

johannesvedder:


core/lib/src/util/supabase_object.dart
```diff
@@ -94,8 +93,12 @@ class SupabaseQuery {
         notExtracted.add(JsonWithError(json, error));
       }
     }
-
-    return ExtractedSupabaseListResult(extracted, notExtracted);
+    if (notExtracted.isNotEmpty) {
+      // If some records could not be extracted, we throw an exception
+      // with the extracted records and the faulty records
+      throw ExtractedSupabaseListResult(extracted, notExtracted);
```
hig-dev, line 99: Rename this to something like `ExtractingFailedException` and make sure that this type `extends Exception`

hig-dev (changes requested):
Using an exception to capture the extracted and not extracted studies is a smart idea and has the benefit that we can leave the other methods unchanged and only change `publishedPublicStudies()`. Personally, I would prefer the solution with less exceptions but that is personal taste.

hig-dev:
This can me merged when #591 is merged.

hig-dev (approved):

PR #586, opened 2024-02-19
Fix/dev bugs 02 2024

johannesvedder:


hig-dev:
@johannesvedder Can you remove the fix android notification icon commit (https://github.com/hpi-studyu/studyu/pull/586/commits/23656f60d3de85d2007e9c88c28656b4503d7e7b)

So that we can use #587 instead.

hig-dev (approved):

PR #587, opened 2024-02-19
fix: missing notification icon

hig-dev:


app/android/app/src/main/res/drawable-xxxhdpi/ic_notification.png
johannesvedder, line 1: xxx-hdpi should have a size of 192x192. See: https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp

johannesvedder (approved):

PR #593, opened 2024-03-15
Fix/offline mode improvements

johannesvedder:
Improves the reliability of the offline mode.

## How to Test this PR

1. Launch a development database and create a study with the Designer.
2. Join a study by using the app in debug mode on a mobile device and skip to the first day.
3. Fill out the observation of the first day, click "next day", and thereafter close (force stop/kill) the app.
4. Shutdown the database.
5. Open the app again. The app should now perform in offline mode. The app state should not have changed.
6. Fill out the observation of the second day. Force the app to stop again.
7. Launch the database again.
8. Start the app. The app state should not have changed and show a tick for the filled out observation of step 6.
9. Validate the entries by logging into the Supabase backend and check if the `subject_progress` table contains the two entries of this test.
10. If any entries are missing or the behavior of the app was different than described here, the test has failed.


hig-dev (approved):
Thanks for the test instructions. It worked.

PR #600, opened 2024-03-26
feat(app): directly show task description in task screen

johannesvedder:


app/lib/screens/study/tasks/task_screen.dart
```diff
@@ -37,16 +37,33 @@ class _TaskScreenState extends State<TaskScreen> {
...
         );
       case QuestionnaireTask questionnaireTask:
-        return QuestionnaireTaskWidget(
-          task: questionnaireTask,
-          key: UniqueKey(),
-          completionPeriod: taskInstance.completionPeriod,
+        return Flex(
+          direction: Axis.vertical,
+          children: [
+            QuestionnaireTaskWidget(
```
hig-dev, line 61: Why do you use a Flex with only one child.

app/lib/screens/study/tasks/task_screen.dart
```diff
@@ -37,16 +37,33 @@ class _TaskScreenState extends State<TaskScreen> {
   Widget _buildTask() {
     switch (taskInstance.task) {
       case CheckmarkTask checkmarkTask:
```
hig-dev, line 39: I was not able to create a CheckmarkTask in the designer. Can you explain me how to do it?

johannesvedder: Any intervention is a `CheckmarkTask`. See: https://github.com/hpi-studyu/studyu/blob/a99ad4b29d8554520b2bf9d89ee8a8a4ace78dbb/designer_v2/lib/features/design/interventions/intervention_task_form_data.dart#L45-L52

app/lib/screens/study/tasks/task_screen.dart
johannesvedder: QuestionnaireTaskWidget is using Expanded that needs to be encapsulated. I refactored the part to make it more clear.

app/lib/screens/study/tasks/observation/questionnaire_task_widget.dart
```diff
@@ -49,43 +49,45 @@ class _QuestionnaireTaskWidgetState extends State<QuestionnaireTaskWidget> {
...
       }),
     );
-    return Expanded(
-      child: Column(
-        children: [
-          Expanded(
-            child: Form(
-              key: formKey,
-              child: questionnaireWidget,
+    return Flex(direction: Axis.vertical, children: [
```
hig-dev, line 52: This Flex is not necessary. I would suggest the following code for the build method:

```
@override
  Widget build(BuildContext context) {
    return Column(
      children: [
        Expanded(
          child: Form(
            key: formKey,
            child: QuestionnaireWidget(
              widget.task.questions.questions,
              header: widget.task.header,
              footer: widget.task.footer,
              onChange: _responseValidator,
              onComplete: (qs) => setState(() {
                response = qs;
              }),
            ),
          ),
        ),
        response != null && responseValidator
            ? ElevatedButton.icon(
                style: ButtonStyle(backgroundColor: MaterialStateProperty.all<Color>(Colors.green)),
                onPressed: () async {
                  if (isRedundantClick(loginClickTime)) {
                    return;
                  }
                  if (!formKey.currentState!.validate()) {
                    return;
                  }
                  setState(() {
                    _isLoading = true;
                  });
                  switch (response) {
                    case QuestionnaireState questionnaireState:
                      await _addQuestionnaireResult<QuestionnaireState>(questionnaireState, context);
                      break;
                  }
                  setState(() {
                    _isLoading = false;
                  });
                },
                icon: _isLoading ? const CircularProgressIndicator(color: Colors.white) : const Icon(Icons.check),
                label: Text(AppLocalizations.of(context)!.complete),
              )
            : const SizedBox.shrink(),
      ],
    );
  }
```

hig-dev (changes requested):

hig-dev (approved):

PR #617, opened 2024-05-13
chore: remove docs action

johannesvedder:
The github action used to create a documentation of all packages is making it difficult to merge branches because of the auto-commit function. It will be helpful to move the documentation to a separate repository, although we might lose the benefit of having multiple documentation versions for each branch.

If the documentation is not used at the moment by anyone, we can move this to an issue.


johannesvedder:
https://github.com/hpi-studyu/studyu/issues/624

PR #622, opened 2024-05-15
fix: UI overflows

ibrahimozkn:
This fix address UI overflows in screens when the system font size is increased. It also resolves the overflow of the text in the report story screen.


johannesvedder (approved):
Nice 🎉 Feel free to merge

PR #644, opened 2024-06-13
fix: duplicate study missing intervention & measurement data

ibrahimozkn:
Fixed the issue where intervention and measurement data is missing when study is being duplicated


johannesvedder (approved):
Tested it and it does the job ✅

PR #655, opened 2024-07-11
fix(designer): Refresh dashboard study table on pin change

johannesvedder:
Designer dashboard study table did not show changes instantly, when studies were pinned/unpinned. This should be fixed now.


ibrahimozkn (approved):
✅

PR #669, opened 2024-07-27
Bug: Monitoring 

rajuAhmed1705:
scope of testing:
https://trello.com/c/yOMuCoYa


johannesvedder:
Is this ready for review? @rajuAhmed1705

rajuAhmed1705:
@johannesvedder, except for the task with the intervention mismatch problem. everything is solved. I actually cannot create a study without an intervention task. So, I'm trying to figure out the other way around.

johannesvedder:
> @johannesvedder, except for the task with the intervention mismatch problem. everything is solved. I actually cannot create a study without an intervention task. So, I'm trying to figure out the other way around.

This used to be possible in the past and we might need to re-enable this again in order to allow washout phases, as long as we have the adaptive schedule alternative. So you can revert [this commit](https://github.com/hpi-studyu/studyu/commit/825ea7fb269412f2afb26bae998288d984167174) if you would like.

ibrahimozkn:
> > @johannesvedder, except for the task with the intervention mismatch problem. everything is solved. I actually cannot create a study without an intervention task. So, I'm trying to figure out the other way around.
> 
> This used to be possible in the past and we might need to re-enable this again in order to allow washout phases, as long as we have the adaptive schedule alternative. So you can revert [this commit](https://github.com/hpi-studyu/studyu/commit/825ea7fb269412f2afb26bae998288d984167174) if you would like.

#671

ibrahimozkn (changes requested):
- UI overflow when language is set to German in bar chart legends and participant details legends
- Missing translation for ""Dropout"

image
image

ibrahimozkn (approved):
✅

PR #670, opened 2024-07-31
fix: unique invite code

rajuAhmed1705:


johannesvedder (changes requested):
- Missing migration file
- We need to handle the case if someone create a duplicate invite code in the Designer. Did you test the validation implemented [here](https://github.com/hpi-studyu/studyu/blob/30835ba33e8b9c1d6a813bc96c1d04c30fe3d0db/designer_v2/lib/features/recruit/invite_code_form_controller.dart#L73)?

johannesvedder:
Review instructions should be:
1. Start local database with this branch
2. Create new study and add an invite code
3. Try to add another invite code, this time use the same invite code as in step 2. It should fail with a distinct error message that describes that the code is already in use.
4. Log in with another user and create a new study and try to add a new invite code with the same name as in step 2. It should fail with the same error message.

johannesvedder:
> Review instructions should be:
> 
>     1. Start local database with this branch
> 
>     2. Create new study and add an invite code
> 
>     3. Try to add another invite code, this time use the same invite code as in step 2. It should fail with a distinct error message that describes that the code is already in use.
> 
>     4. Log in with another user and create a new study and try to add a new invite code with the same name as in step 2. It should fail with the same error message.

The last test case (4) fails for me. The database check works, but the error is not handled by the UI

grafik

rajuAhmed1705:
@johannesvedder Yes please take over the PR. I have not implemented postgres functions before. This will take long time for me.

johannesvedder:
@rajuAhmed1705 please check

johannesvedder (approved):

PR #676, opened 2024-08-08
fix: Make invite codes unique

johannesvedder:
Successor of https://github.com/hpi-studyu/studyu/pull/670

This PR makes sure that invite codes are unique and cannot be used by multiple studies.

## Review instructions

1. Start local database with this branch
2. Create new study and add an invite code
3. Try to add another invite code, this time use the same invite code as in step 2. It should fail with a distinct error message that describes that the code is already in use.
4. Log in with another user and create a new study and try to add a new invite code with the same name as in step 2. It should fail with the same error message.


ibrahimozkn (approved):
- [X] Start local database with this branch
- [X] Create new study and add an invite code
- [X] Try to add another invite code, this time use the same invite code as in step 2. It should fail with a distinct error message that describes that the code is already in use.
- [x] Log in with another user and create a new study and try to add a new invite code with the same name as in step 2. It should fail with the same error message.

Migration file works as well ✅

PR #680, opened 2024-08-25
Feature/filter

rajuAhmed1705:


ibrahimozkn (changes requested):
1. After changing any filter, if the resulting study list is empty, it becomes impossible to access the filtering options to revert the applied changes.

image


2. Choosing both Everyone and Invite Only options in participation filter is duplicating the existing studies in the list
image

ibrahimozkn (dismissed):

johannesvedder:
@ibrahimozkn could you check the warnings? https://github.com/hpi-studyu/studyu/actions/runs/10976108923/job/30476497785?pr=680

johannesvedder:
The e2e test seems to be broken, not sure if this is related to this PR though

johannesvedder:
Some remarks:

- I would still show the sorting arrow next to the name and then the filter icon. Because currently you don't expect that you can still sort the items and only see the filter icon. But the filter menu does not open when you click on the name, even though the icon hover effect appears. So for now I would remove the onclick from the column name, add sorting arrow and filtering icon and have onclick actions with hover effect on both icons.
  - For long term this will probably become really messy to show two symbols next to each column. I like supabase's approach to have a single sort and filter option on the top for all columns. This looks way cleaner.
grafik
- We should think about storing sorting and filtering selections in the database for each user to make these settings persistent
- Since you can only filter by one item right now, radio buttons are a better fit than checkboxes which indicates that a user can select multiple items
- Replace the button label "Filter" with a clearer message. "Apply filter" or just "Apply" fits better 
grafik
- When the filter menu is shown, no other menus/popups/buttons of the page behind should be clickable. A click next to the menu should close the popup. Otherwise this leads to a weird behavior, e.g. when you click on a study while the popup is open: 
grafik

PR #683, opened 2024-09-04
fix: App Debug Screen & Result history

johannesvedder:
- made debug screen more reliable against exceptions
- result history only shows entries from own study_subjects


johannesvedder:
The E2E tests seem to fail, but idk what's the reason. We can probably ignore it for now

ibrahimozkn (approved):

PR #696, opened 2024-10-25
fix: exclude notification and battery optimization debug options on web

ibrahimozkn:


johannesvedder:
Can this be closed @ibrahimozkn ?

PR #720, opened 2025-06-02
Fix/loading screen loop

johannesvedder:


app/lib/screens/app_onboarding/loading_screen.dart
```diff
@@ -105,7 +114,37 @@ class _LoadingScreenState extends State<LoadingScreen> {
...
+                ],
+              );
+            },
+          );
+          if (result == true) {
+            await SecureStorage.deleteAll();
+            StudyULogger.info("Secure storage data deleted");
+          } else {
+            StudyULogger.info("User chose not to delete secure storage data");
+          }
```
johannesvedder, line 147: Initially, when re-installing the app, due to a bug in flutter secure storage the storage might have been unreadable, due to a BadPaddingException. Other causes such as switching the database with development versions or other unknown reasons might also lead to this point, where an Exception is thrown but the user is not getting redirected and is stuck in the loading screen. With the new prompt, the user may reset the secure storage and start over.

flutter_common/lib/src/utils/storage.dart
```diff
@@ -35,7 +36,11 @@ class SupabaseStorage extends LocalStorage {
 }
 
 class SecureStorage {
-  static const storage = FlutterSecureStorage();
+  static const storage = FlutterSecureStorage(
+    aOptions: AndroidOptions(
+      encryptedSharedPreferences: true,
+    ),
+  );
```
johannesvedder, line 43: Apparently, there was no encryption enabled for Android without this setting.

flutter_common/lib/src/utils/storage.dart
```diff
@@ -51,16 +56,26 @@ class SecureStorage {
...
+        StudyULogger.error(
+          "Error reading key $key from secure storage: $e",
+        );
+        if (e is PlatformException && e.code == 'BadPaddingException') {
+          StudyULogger.error(
+            "BadPaddingException: $e. This might indicate that the secure storage is corrupted.",
+          );
+        }
+        rethrow;
+      }
```
johannesvedder, line 71: This is just for logging purposes

flutter_common/lib/src/utils/storage.dart
```diff
@@ -51,16 +56,26 @@ class SecureStorage {
...
 
   static Future<bool?> readBool(String key) async {
-    return await storageLock.synchronized(() async {
-      final readValue = await storage.read(key: key);
-      if (readValue == null) return null;
-      return bool.parse(readValue);
-    });
+    final readValue = await read(key);
+    if (readValue == null) return null;
+    return bool.parse(readValue);
```
johannesvedder, line 78: Uses the original read() method now to log any errors

ibrahimozkn (approved):

PR #699, opened 2024-11-18
feat(designer): reset feature on test mode now resets all subject data

ibrahimozkn:
Reset feature on test mode in designer now resets all subject data that belongs to the editor.

PR #725, opened 2025-06-05
feat: reset test preview on navigate test page

ibrahimozkn:
- Automatically reset test preview when navigating to the test page in the designer.


johannesvedder:
This doesn't work unless https://github.com/hpi-studyu/studyu/pull/699 is fixed, right?

PR #741, opened 2025-09-09
feat: add debug mode functionality for study progression

johannesvedder:


app/lib/screens/study/dashboard/dashboard.dart
```diff
@@ -40,7 +41,9 @@ class _DashboardScreenState extends State<DashboardScreen>
   List<TaskInstance>? scheduleToday;
 
   bool get showNextDay =>
-      (kDebugMode || context.read<AppState>().isPreview) &&
+      (kDebugMode ||
+          context.read<AppState>().isPreview ||
+          context.read<AppState>().isDebugModeEnabled) &&
```
johannesvedder, line 46: Having different "debug" variables is also not optimal, can become confusing. Maybe we can combine all of them and use them in the whole application?
So we can have:

`debugMode = isPreview || kDebugMode`

And use `debugMode` everywhere instead of `kDebugMode`. What do you think?

app/lib/screens/study/dashboard/settings.dart
```diff
@@ -93,53 +95,145 @@ class _SettingsState extends State<Settings> {
...
-                  builder: (_) => DeleteAlertDialog(subject: subject),
-                );
-              },
+            const SizedBox(height: 8),
+            Text(
+              'Tap the settings title 7 times to show/hide these controls',
+              style: Theme.of(context).textTheme.bodySmall?.copyWith(
+                color: Colors.grey[600],
+                fontStyle: FontStyle.italic,
+              ),
```
johannesvedder, line 166: We already have a hidden debug menu, so we can just add a switch there. No need for another hiding place

johannesvedder (commented):
Maybe let's also have a DEBUG text in the dashboard somewhere if its not taking away too much space?!

johannesvedder:
I work on an easier alternative myself

PR #744, opened 2025-09-10
fix: study duplication data loss

johannesvedder:
All data stored in lists (screening, consent, interventions, measurements, reports) is lost afterwards (not only visual but in database) and cannot be restored anymore.

Some additional finding: when copying a study as draft in the registry, the study list switches to my studies, but the sidebar does not reload

Steps to reproduce:

- Clone a study (from study registry or from my studies)
- Open the new study
- Go back to dashboard
- Open the new study again
- Perform a change in the study info section (maybe changing some other field would also work)
- All data in the list fields disappear and the broken study is pushed to server.

----

## Additional changes

- When duplicating a study, the study is directly opened.


ibrahimozkn (approved):
🎉

PR #795, opened 2026-02-09
fix: update free text validation regex handling

ibrahimozkn:
This pull request improves the validation and user experience for custom free-text questions by simplifying error messages, ensuring consistent regex validation, and refactoring related code in both the app and the designer. The changes focus on making custom input validation more robust and user-friendly, especially for fields requiring custom regular expressions.

**Localization and Error Message Simplification:**

- Changed the `free_text_custom_error` localization in both English and German to a simpler, more general message, and removed the `{pattern}` placeholder from all relevant files (`app_en.arb`, `app_de.arb`, `app_localizations.dart`, `app_localizations_en.dart`, `app_localizations_de.dart`). [[1]](diffhunk://#diff-73633819f5579092c04c0f8fea317f5cfb4f86eb2826f887634a006cdf53c9afL233-R233) [[2]](diffhunk://#diff-2cbc3c3416c201eed4e085628be20a938df8890fa72c4896cb14c987decc3f5fL237-R237) [[3]](diffhunk://#diff-d89db88900d71e4fb1fbd51fb15a0111ae237cc733dc718fdc583eb46b2fd8c6L1178-R1179) [[4]](diffhunk://#diff-6ebaa03203282a8ccdd1bdd574159a113ddc50266c2c97598d6cd52caa9a1dc7L599-R600) [[5]](diffhunk://#diff-ab5c9921c4f0ab6251b1ed77e5880faf4861ee91e50e09f335b46652ff50f7f7L606-R607)

**App-Side Free-Text Validation Refactor:**

- Updated the validation logic in `free_text_question_widget.dart` to use a new helper (`free_text_regex_validation.dart`) for building and validating full-match regex patterns, and to use the simplified error message when validation fails. [[1]](diffhunk://#diff-702c7622b8b792bbc3b8fd08485183302adca1ba4df82ddd3d6a157103b57eeeR6) [[2]](diffhunk://#diff-702c7622b8b792bbc3b8fd08485183302adca1ba4df82ddd3d6a157103b57eeeR166-R171) [[3]](diffhunk://#diff-702c7622b8b792bbc3b8fd08485183302adca1ba4df82ddd3d6a157103b57eeeL202-R212) [[4]](diffhunk://#diff-f6ccc8d999afeaa374627c0efcc22e5da7b85cdbbecd7f565b8a862a925805f2R1-R20)

**Designer-Side Regex Validation Improvements:**

- Enhanced the `QuestionFormViewModel` in the designer to validate custom regex expressions more robustly, ensuring invalid patterns are caught early, and to update related controls when the free-text type changes. [[1]](diffhunk://#diff-f4a6e03029849ad5bd768f0af48fb1297bbbf3ed802d4b8c88ee74eb7ba3f9d8R36-R43) [[2]](diffhunk://#diff-f4a6e03029849ad5bd768f0af48fb1297bbbf3ed802d4b8c88ee74eb7ba3f9d8R380-R417) [[3]](diffhunk://#diff-f4a6e03029849ad5bd768f0af48fb1297bbbf3ed802d4b8c88ee74eb7ba3f9d8L384-R434)
- Updated the free-text question form view to display appropriate validation messages for required fields and invalid patterns.


johannesvedder (changes requested):
1. Complete is still possible ven though text is not in required format. Is this fixable?

Image

designer_v2/lib/features/design/shared/questionnaire/question/question_form_controller.dart
```diff
@@ -369,6 +377,44 @@ class QuestionFormViewModel extends ManagedFormViewModel<QuestionFormData>
         validators: [Validators.delegate(_validateFreeText)],
       );
 
+  void _onFreeTextTypeChanged(FreeTextQuestionType? _) {
```
johannesvedder, line 381: Can we disable "allowed range of text length" if Custom is specified as allowed text type. For example in this case regular expression wants to have five chars, but the allowed range of text length is set to 0-3 chars so it will never match since both are overlapping. If user chooses custom they should be able to control format and length of the input and have full control over all with regex in power mode.

Image

johannesvedder (changes requested):

johannesvedder:
Previous bug is still not fixed. Numeric input only allows empty values
image

app/lib/widgets/questionnaire/questions/free_text_question_widget.dart
```diff
@@ -90,6 +94,12 @@ class _FreeTextQuestionWidgetState extends State<FreeTextQuestionWidget> {
     if (_formFieldKey.currentState?.validate() == true) {
       widget.onDone?.call(widget.question.constructAnswer(value));
       _hasSubmitted = true;
+      _hadValidSubmission = true;
+    } else if (_hadValidSubmission) {
+      // If we had a valid submission before but now it's invalid,
+      // notify the parent to remove this answer
+      widget.onInvalid?.call();
+      _hadValidSubmission = false;
     }
```
johannesvedder: I don't really like this solution, why not stick with validation messages as we had before? What's the reason for the change?

Current problem for example let's say the validation criteria is alphanumeric and the user types some text and adds a colon or a dot all the previous text will be removed without a single error message.

johannesvedder (changes requested):

ibrahimozkn:
> Previous bug is still not fixed. Numeric input only allows empty values image

Did you set min length to 1 or is it 0 in the designer?

ibrahimozkn:
@johannesvedder 

- Removed the onInvalid callback pattern validation now works through error messages shown directly under the input field, as you suggested
- Fixed the bug where typing invalid characters (like a colon in an alphanumeric field) would silently delete all previously entered text the text now stays and an error message explains what's wrong
- Removed the Submit button, validation and answer submission now happen automatically as the user types, so the questionnaire state is always up to date
- If a user changes a valid answer to something invalid, the answer is automatically cleared and the Complete button won't appear until the input is corrected
- Added a new "Error message" field in the designer for custom regex questions, researchers can now write a human-readable error message (e.g. "Please enter a code in format: XX-1234") instead of participants seeing a generic "Please enter a value in the required format"
- If no custom error message is set, the generic message is used as fallback

app/lib/widgets/questionnaire/questions/free_text_question_widget.dart
```diff
@@ -135,10 +97,9 @@ class _FreeTextQuestionWidgetState extends State<FreeTextQuestionWidget> {
   List<TextInputFormatter> _getInputFormatters() {
     switch (widget.question.textType) {
       case FreeTextQuestionType.numeric:
-        return [FilteringTextInputFormatter.allow(RegExp('^-?[0-9]*'))];
-      case FreeTextQuestionType.alphanumeric:
-        return [FilteringTextInputFormatter.allow(RegExp(alphanumericPattern))];
+        return [FilteringTextInputFormatter.allow(RegExp('[0-9-]'))];
```
johannesvedder: This will cause different behavior with existing study setups.

johannesvedder: e.g It will allow the user to type 12-34-56 or ---99, which are not valid numbers.

johannesvedder (changes requested):
todo:
- [ ] check what is actually broken and is supposed to be fixed by this PR. Only address the necessary changes and avoid to include the overengineered AI code.

PR #798, opened 2026-02-23
fix: dashboard column sorting

ibrahimozkn:
The Equatable mixin in DashboardState was configured with an incomplete props list. It only included [studies, studiesFilter] but excluded sortByColumn, sortAscending, and query.                                                                                                                                                    
                                         
Riverpod uses == to compare states (powered by Equatable). When setSorting() updated the sort fields, the incomplete props list caused the comparison to return true (states appear equal), so Riverpod discarded the update and the UI never rebuilt.

Added all state fields that affect the displayed output to props:
  - sortByColumn - which column is being sorted
  - sortAscending - the sort direction
  - query - the search filter

Now when users click a column header:
  1. setSorting() updates the state with new sort parameters
  2. Riverpod detects the state change (since props now includes sort fields)
  3. The UI rebuilds and calls displayedStudies() with the new sort parameters
  4. The table displays correctly sorted data


johannesvedder (approved):

johannesvedder:
Not sure why E2E is failing, but if it might fail in other PRs later on we need to look into it

PR #806, opened 2026-03-23
fix(database): restore cascade delete for study dependencies

ibrahimozkn:
## Summary
- restore `ON DELETE CASCADE` on the `study` foreign keys used by `repo`, `study_fitbit_credentials`, `study_invite`, and `study_subject`
- add a forward migration to safely update existing instances
- align `database/studyu-schema.sql` and add a regression test covering study deletion cleanup

### Self-hosted Supabase
1. Start the local Supabase stack:
   ```bash
   supabase start
   ```
2. Apply the latest database state from this branch on a fresh local instance:
   ```bash
   supabase db reset
   ```
3. Run the SQL test suite, including the new cascade regression coverage:
   ```bash
   supabase test db
   ```

### Expected result
- deleting a row from `public.study` also deletes dependent rows in `public.repo`, `public.study_fitbit_credentials`, `public.study_invite`, and `public.study_subject`
- the regression test in `supabase/tests/010-study.sql` passes

### DIFF
--- a/database/migration/20260323_restore_study_delete_cascade.sql
+++ b/database/migration/20260323_restore_study_delete_cascade.sql
@@ -1,0 +1,36 @@
+-- ============================================================================
+-- Migration: Restore Study Delete Cascade
+-- Date: March 23, 2026
+-- ============================================================================
+--
+-- OVERVIEW:
+-- Restore ON DELETE CASCADE behavior for foreign keys that reference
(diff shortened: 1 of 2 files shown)


johannesvedder:
The description you added will not work by default when using the seed.sql Use seed-ci.sql for testing to let the supabase tests pass.
with seed.sql
image
With seed-ci.sql as seed.sql
image

johannesvedder (approved):

PR #807, opened 2026-04-04
Improve study test iframe preview health checks, app status messaging, and overlays

mohiuddinshahrukh:
## Summary

This PR improves the study test preview flow between `designer_v2` and the embedded app iframe.

It adds:
- app reachability checks before attempting iframe preview
- staged preview overlay states in the designer
- app-to-designer preview status messaging for loading/loaded/error
- nicer glass-style preview overlays
- safer preview handling for incomplete studies
- a fix so the selected preview route is not reset during rebuilds

## Behavior

The test preview now follows this flow:
- health check
- iframe connection
- app loading
- app shown

If the app is unreachable:
- local/dev shows a local app unavailable message
- deployed mode shows a maintenance/unavailable message

## Scope

This PR intentionally includes only the files related to:
- iframe preview health/status handling
- preview overlay UX
- preview route/reset reliability


johannesvedder (changes requested):
Thanks for the PR and the additions. It looks much better now! Some things I noticed:

1. Can we remove the transparency? image

2. Let's skip the /onboarding for the preview as it's purpose is to introduce new participants to the app: 

Image

3. When switching pakes using the links on the left side, I am stuck in the checking connection loading screen. I need to click "Reset" to show the correct page.

mohiuddinshahrukh:
One two and three have been fixed, although number 3 is a bit more complicated to solve since the back button action is a bit of a tricky thing to handle, it sometimes makes the phone app crash and sometimes behaves unexpectedly, this can be easily reproduced by clicking schedule and then back on the deployed version also.

Shall you find some other room for improvements, let me know.

johannesvedder (changes requested):
praise: Thanks this works great now. Some small improvement ideas that we could maybe include into this PR:

1. I think on first load the Study Overview should be shown. Displaying the app start does not make sense when previewing a specific study.

Image

2. In case there are no screening questions, the screener in the sidebar should be disabled.

Image

3. I feel like we could increase the height/width of the phone widget to improve readability in general. Can we just increase the current settings and will this work with different screen sizes?

Image

mohiuddinshahrukh:
praise: Thanks for the feedback on the PR, these changes make the system feel more userfriendly and are helping use the system in a more controlled way!

All expected changes have been made 1,2 and 3.
For 3 I did the following: 

- Phone target size 300x600 -> 360x720.
- Added minWidth = 260.
- Preview frame now uses available constraints and FittedBox(scaleDown) so it stays usable on smaller screens. Simply said - its responsive.

You may review again!

PR #808, opened 2026-04-07
Fix/designer snackbar auto dismiss

mohiuddinshahrukh:
## Summary

This PR restores the previous auto-dismiss behavior for designer snackbars.

## Root cause

The shared snackbar dispatcher always adds an action button (`X`).
In the current Flutter SDK, snackbars with an action default to `persist = true`,
which caused snackbar notifications to remain visible until manually dismissed.

## Fix

Set `persist: false` in the shared snackbar builder so actionable snackbars
continue to auto-dismiss after their configured duration.

## Affected behavior

This fixes shared snackbar notifications such as:
- wrong password / invalid credentials messages
- delete study confirmation feedback
- other designer snackbar messages routed through the notification dispatcher

### DIFF
--- a/app/lib/screens/app_onboarding/iframe_helper.dart
+++ b/app/lib/screens/app_onboarding/iframe_helper.dart
@@ -1,25 +1,43 @@
 import 'dart:convert';
 
 import 'package:studyu_app/models/app_state.dart';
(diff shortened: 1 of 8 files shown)


mohiuddinshahrukh:
Contaminated with previous commits.

PR #809, opened 2026-04-07
Add sortable serial column to studies table

mohiuddinshahrukh:
## Summary

This PR adds a serial column to the studies table and makes it sortable.

## Changes

- add a dedicated `Serial` column to the studies table
- display stable serial numbers for visible studies
- allow sorting by the serial column
- keep the displayed serial values aligned with the serial sort order

## Notes

The serial value is a table-level ordering aid for the studies list.
It is computed from study creation order for the currently visible study set.

image

### DIFF
--- a/app/lib/screens/app_onboarding/iframe_helper.dart
+++ b/app/lib/screens/app_onboarding/iframe_helper.dart
@@ -1,25 +1,43 @@
 import 'dart:convert';
 
 import 'package:studyu_app/models/app_state.dart';
-import 'package:studyu_core/core.dart';
+import 'package:studyu_core/core.dart' show Study;
 import 'package:studyu_core/env.dart' as env;
 import "package:universal_html/html.dart" as html;
 
 class IFrameHelper {
+  String _designerOrigin() {
+    final referrer = html.document.referrer;
+    if (referrer.isNotEmpty) {
+      final uri = Uri.tryParse(referrer);
+      if (uri != null && uri.hasScheme && uri.host.isNotEmpty) {
+        return uri.origin;
+      }
+    }
+    return env.designerUrl!;
+  }
+
+  void postPreviewStatus({required String status, String? message}) {
+    html.window.parent!.postMessage(
+      jsonEncode({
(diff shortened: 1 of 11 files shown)


mohiuddinshahrukh:
Contaminated with previous commits.

johannesvedder:
> Contaminated with previous commits.

I think the live pinning functionality got broken by this PR (or has been broken before), as the pinning is only shown after reloading the page.

Interesting idea in general. Are stable IDs in dashboards something other tools provide as well or is this giving researchers a specific advantage? I think you already recognize your studies well enough using their name, so I am unsure if this is necessary to include. Moreover, naming the column "ID" does not make it clear what it is supposed to do: Users read that column as an immutable ID, but it behaves like a derived row order, since when deleting study with ID 1, study with ID 2 will replace ID 1, etc. Additionally, it is not purely positional, since IDs don't change with sort/filter/search. I find it more confusing than helpful, so I would not merge it right now. Another contra argument is that for small screens the horizontal space is already limited and the ID column takes away more space.
What do you think?

mohiuddinshahrukh:
Thank you for going through the PR.
Although the inclusion of a serial # column is not necessary, here is some experience I have with a few systems I have worked with...


Regarding your feedback: 

1. ID is perhaps not the right name to use here, it should rather be serial #, i agree with that.
2. The idea of this is to have to not count the studies, if we think of studyU from a users perspective, we should offer this as it is a convenience and a testers perpective, these are basic requirements of a system, since there is also a lack of pagination the system has a page that can be as long as the browser allows, but perhaps this is a bug to discuss another time.
3. Side info: From a system perspective, we should never expose original IDs to users 
4. For smaller screens I think the table should be resizable, regardless of if we keep serial # column or not, or there should be a better solution for this.
5. Pinning has been broken from before, need to take a look at that too.


Let me know your thoughts too please and we will proceed accordingly!

johannesvedder:
With the upcoming pagination feature I think we can close this. Let's include the number of studies in the list in that PR!

PR #811, opened 2026-04-07
Add sortable serial column to studies table

mohiuddinshahrukh:
## Summary

This PR adds a serial column to the studies table and makes it sortable.

## Changes

- add a dedicated `Serial` column to the studies table
- display stable serial numbers for visible studies
- allow sorting by the serial column
- keep the displayed serial values aligned with the serial sort order

## Notes

The serial value is a table-level ordering aid for the studies list.
It is computed from study creation order for the currently visible study set.

image


johannesvedder:
Duplicate of #809? Can we close this?

PR #812, opened 2026-04-07
Improve studies table header typography

mohiuddinshahrukh:
## Summary

Adjust studies table header typography for better readability and hierarchy.

## Changes

- use `labelMedium` for header labels
- increase header text weight (`w700`)
- slightly increase contrast
- add subtle letter spacing for clearer scanning

## Scope

Only updates studies table header presentation.
No behavior, sorting, or data logic changes.

image

### DIFF
--- a/designer_v2/lib/features/dashboard/studies_table_column_header.dart
+++ b/designer_v2/lib/features/dashboard/studies_table_column_header.dart
@@ -31,6 +31,12 @@ class _StudiesTableColumnHeaderState extends State<StudiesTableColumnHeader> {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final headerTextStyle =
+        (theme.textTheme.labelLarge ?? theme.textTheme.bodyMedium)!.copyWith(
+          fontWeight: FontWeight.w700,
+          color: theme.colorScheme.onSurface.withValues(alpha: 0.88),
+          letterSpacing: 0.15,
+        );
 
     return MouseEventsRegion(
       key: ValueKey('studies_table_column_header_${widget.title}'),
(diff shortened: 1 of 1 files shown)


johannesvedder (commented):
Thanks! I think the font size is too little with labelMedium if you compare it. Makes it difficult to read and reduces accessibility.

Image

Image

mohiuddinshahrukh:
Dear Johannes, the font has been changed to labelLarge with a bodyMedium fallback.
Should be more visible now!

designer_v2/lib/features/dashboard/studies_table_column_header.dart
```diff
@@ -31,6 +31,12 @@ class _StudiesTableColumnHeaderState extends State<StudiesTableColumnHeader> {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final headerTextStyle =
+        (theme.textTheme.labelLarge ?? theme.textTheme.bodyMedium)!.copyWith(
+          fontWeight: FontWeight.w700,
+          color: theme.colorScheme.onSurface.withValues(alpha: 0.88),
+          letterSpacing: 0.15,
+        );
```
johannesvedder, line 39: I just tested it again:

- labelLarge is always present in Material themes. The ?? theme.textTheme.bodyMedium is unreachabable and unnecessary, no?
- The heading labels are still smaller than what we had originally, so this doesn't increase readability imho. The changed fontWeight: FontWeight.w700 (bold for all is too aggressive) and letterSpacing: 0.15 (kind of arbitrary, default is 0.5) on all headers are also not following M3 defaults. I would recommend to stick with using our customized bodySmall since this is used a lot of times in the project to keep everything aligned and close this PR. Alternatively, we could change it to labelLarge without the non-conform m3 customization to keep the same font size, but make a fix based on our theme and not on an individual basis, but this seems like an overkill for this change.

johannesvedder (commented):
suggestion: See analysis in PR thread

designer_v2/lib/features/dashboard/studies_table_column_header.dart
mohiuddinshahrukh: I see, yes in that case, I'll close this PR.

PR #813, opened 2026-04-07
Polish studies search bar spacing and alignment in dashboard

mohiuddinshahrukh:
### Summary
Improves the studies dashboard search input styling so it aligns better with adjacent controls and looks cleaner in Material 3 UI.

### Changes
- Updated shared search component styling in `designer_v2/lib/common_views/search.dart`:
  - Added consistent internal padding.
  - Refined leading icon spacing and icon style.
  - Applied subtle M3-aligned surface/background and outline styling.
  - Set compact fixed input height (`35`) to align with nearby filter control.
  - Kept flat elevation/shadow behavior.
- Updated studies page search hint text to `Search` in `designer_v2/lib/features/dashboard/dashboard_page.dart`.

### Scope
- UI-only change.
- No logic, API, or data behavior changes.

### Validation
- Verified visually on the `/studies` dashboard:
  - Search field is aligned with the filter button.
  - Input feels less cramped and more consistent.
  - Hint text appears as `Search`.

### DIFF
--- a/designer_v2/lib/common_views/search.dart
+++ b/designer_v2/lib/common_views/search.dart
@@ -56,17 +56,9 @@ class SearchState extends State<Search> {
         child: Icon(Icons.search_rounded, size: 20),
       ),
       elevation: const WidgetStatePropertyAll<double>(0),
-      constraints: const BoxConstraints(
-        minHeight: 35,
-        maxHeight: 35,
-      ),
-      backgroundColor: WidgetStatePropertyAll<Color>(
-        theme.colorScheme.surfaceContainerHighest.withValues(alpha: 0.45),
-      ),
+      constraints: const BoxConstraints(minHeight: 35, maxHeight: 35),
       side: WidgetStatePropertyAll<BorderSide>(
-        BorderSide(
-          color: theme.colorScheme.outlineVariant,
-        ),
+        BorderSide(color: theme.colorScheme.outlineVariant),
       ),
       padding: const WidgetStatePropertyAll<EdgeInsetsGeometry>(
         EdgeInsetsDirectional.fromSTEB(12, 0, 12, 0),
--- a/designer_v2/lib/common_views/search.dart
+++ b/designer_v2/lib/common_views/search.dart
@@ -47,10 +47,36 @@ class SearchState extends State<Search> {
 
   @override
   Widget build(BuildContext context) {
+    final theme = Theme.of(context);
     return SearchBar(
       hintText: widget.hintText ?? "Search",
       controller: _searchController,
-      leading: const Icon(Icons.search),
+      leading: const Padding(
+        padding: EdgeInsetsDirectional.only(start: 4),
+        child: Icon(Icons.search_rounded, size: 20),
+      ),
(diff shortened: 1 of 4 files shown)


johannesvedder (commented):
Great thanks, looks much cleaner now. Just a question: What was the reason for the removed localization tr.search? Can we restore this?

mohiuddinshahrukh:
Thanks for reviewing!
I restored tr.search as that was a mistake on my end, I misunderstood its purpose!

designer_v2/lib/common_views/search.dart
```diff
@@ -47,10 +47,36 @@ class SearchState extends State<Search> {
...
     return SearchBar(
       hintText: widget.hintText ?? "Search",
       controller: _searchController,
-      leading: const Icon(Icons.search),
+      leading: const Padding(
+        padding: EdgeInsetsDirectional.only(start: 4),
+        child: Icon(Icons.search_rounded, size: 20),
+      ),
+      elevation: const WidgetStatePropertyAll<double>(0),
+      constraints: const BoxConstraints(
```
johannesvedder: Remove this `backgroundColor` override and let Flutter's M3 default handle it.

The original `SearchBar` had no `backgroundColor` set, which means Flutter uses `colorScheme.surfaceContainerHigh` — solid, fully opaque, and adapts correctly to both light and dark themes.

The current `surfaceContainerHighest.withValues(alpha: 0.45)` is semi-transparent, which means it bleeds through the scaffold background color and looks different depending on the surface it sits on.

```dart
// remove this block:
backgroundColor: WidgetStatePropertyAll<Color>(
  theme.colorScheme.surfaceContainerHighest.withValues(alpha: 0.45),
),
```

Flutter's `SearchBar` default (`surfaceContainerHigh`, fully opaque) is the solid light-gray background visible in the original screenshot — the correct M3 appearance.

johannesvedder (changes requested):
Overall the changes look good — consistent use of M3 color tokens, matching border treatment with the Filter button, flat elevation, and correct typography. One change requested on the `backgroundColor` override.

designer_v2/lib/common_views/search.dart
mohiuddinshahrukh: Removed.

designer_v2/lib/features/app_controller.dart
```diff
@@ -37,9 +37,9 @@ class AppController extends _$AppController {
       (delegate) => delegate.onAppStart(),
       withMinDelay: true,
     );
-    _stateController.add(
-      const AppControllerState(status: AppStatus.initialized),
-    );
+    const initializedState = AppControllerState(status: AppStatus.initialized);
+    state = const AsyncValue.data(initializedState);
+    _stateController.add(initializedState);
```
johannesvedder: How did this end up here? Redundant state assignment before stream emission. I think this is safe to revert?

johannesvedder (changes requested):
praise: Looks much better now!

johannesvedder (approved):

PR #814, opened 2026-04-07
Add `Actions` header for the studies table action column

mohiuddinshahrukh:
### Summary
Adds a visible header label for the last column in the studies table so the column purpose is clear.

### Changes
- Updated `designer_v2/lib/features/dashboard/studies_table.dart`.
- Set the `StudiesTableColumn.action` header title to `Actions`.
- Kept the action column non-sortable (no sorting behavior added).

### Scope
- UI-only table header labeling change.
- No data, backend, or sorting logic behavior changes.

### Validation
- Verified the studies table now shows `Actions` as the last column header.
- Verified the `Actions` column remains non-sortable.

### DIFF
--- a/designer_v2/lib/features/dashboard/studies_table.dart
+++ b/designer_v2/lib/features/dashboard/studies_table.dart
@@ -281,8 +281,9 @@ class StudiesTable extends StatelessWidget {
       case StudiesTableColumn.completed:
         title = tr.studies_list_header_participants_completed;
       case StudiesTableColumn.pin:
-      case StudiesTableColumn.action:
         title = '';
+      case StudiesTableColumn.action:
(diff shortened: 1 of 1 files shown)


johannesvedder (commented):
Thanks for the idea. I think  if the column only contains an ellipsis menu and is not sortable/filterable, we can leave the header visually empty to reduce noise. We still have a tooltip for the menu to provide explanation, so this keeps the table clean without losing clarity

mohiuddinshahrukh:
Thanks for the feedback.

The intention behind this was to follow convention: If all columns have headers, then this should also follow the convention, it makes the table look more aligned but I guess this can be taken as food for thought and considered later on aswell to be something included or left out!

PR #815, opened 2026-04-07
Adjust New Study button icon spacing and compact sizing

mohiuddinshahrukh:
### Summary
Refines the `New study` button on the dashboard toolbar to improve visual balance and spacing while keeping the overall button compact.

### Changes
- Updated `designer_v2/lib/common_views/primary_button.dart`:
  - Added configurable `iconTextSpacing` parameter to `PrimaryButton`.
  - Replaced fixed icon-label gap with `iconTextSpacing`.
- Updated `designer_v2/lib/features/dashboard/dashboard_page.dart`:
  - Applied compact toolbar button sizing for `New study`:
    - `height: 30`
    - `minimumSize: Size(0, 30)`
    - symmetric inner padding (`horizontal: 6`, `vertical: 0`)
  - Increased icon/text distance specifically for this button with:
    - `iconTextSpacing: 14`

### Scope
- UI-only toolbar button polish.
- No behavior or backend changes.

### Validation
- Verified `New study` button remains compact.
- Verified icon and label have clearer separation.
- Verified button remains aligned with surrounding toolbar elements.

### DIFF
--- a/designer_v2/lib/common_views/primary_button.dart
+++ b/designer_v2/lib/common_views/primary_button.dart
@@ -16,6 +16,7 @@ class PrimaryButton extends StatefulWidget {
       horizontal: 4.0,
       vertical: 8.0,
     ),
+    this.iconTextSpacing = 6.0,
     this.minimumSize,
     super.key,
   });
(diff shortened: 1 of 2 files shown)


johannesvedder (commented):
Old
Image

New
Image

I prefer the old and more rounded button more than the new stretched one to be honest.

mohiuddinshahrukh:
Thank you for the review.

My take on this would be that to make the UI look aligned we either change all the buttons to be following one convention, E.g the filter button looks different from the new study button, (as it should but there should also be some consistency in the system) but the sizes and spacing between icons and texts should be consistent etc.

What do you think about that?

PR #817, opened 2026-04-22
fix(dashboard): restore studies table pinning

mohiuddinshahrukh:
Move pinned study state into the dashboard controller so table pin/unpin updates reactively, and use the current toggle state when applying pin actions.

### DIFF
--- a/designer_v2/lib/features/dashboard/dashboard_controller.dart
+++ b/designer_v2/lib/features/dashboard/dashboard_controller.dart
@@ -71,6 +71,7 @@ class DashboardController extends _$DashboardController
       state = state.copyWith(
         savedFilters: () => savedFilters,
         activeFilter: () => active.filterGroup,
+        pinnedStudies: () => _userRepository.user.preferences.pinnedStudies,
         selectedSavedFilterId: () => active.presetId,
       );
     } catch (e) {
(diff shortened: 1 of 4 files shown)


johannesvedder:
Does this coincide with https://github.com/hpi-studyu/studyu/pull/840?

mohiuddinshahrukh:
Do you mean, if that has been fixed in PR 840?
Then no, coincidentally i tested it :P in PR 840 [infinite scrolling one] and also in the video you can see that the pinning is still not visible, maybe @ibrahimozkn could also let us know?

johannesvedder:
> Do you mean, if that has been fixed in PR 840? Then no, coincidentally i tested it :P in PR 840 [infinite scrolling one] and also in the video you can see that the pinning is still not visible, maybe @ibrahimozkn could also let us know?

There are a few points that need to be validated with #840: We must verify infinite scrolling doesn't interfere with pinned studies, e.g. pinned studies should always load first in the study list so they need to be prioritized on server side. Filtering + sorting is now on server side so pinning should be too I guess? Also check if there are other edge cases we need to take care of e.g. what happens if there are more pinned studies than the initially loaded number of studies and so on.

mohiuddinshahrukh:
@johannesvedder Shall we close this one aswell then since it's going into something to be considered for PR #840 ?

PR #818, opened 2026-04-22
fix(app): clarify study selection banner message

mohiuddinshahrukh:
## What changed
Updates the study selection warning banner copy to use neutral, user-friendly wording.

## Why
The banner is shown when some study records fail to deserialize, but the previous copy told users their app version was outdated. That message overstated the cause and could be misleading when the actual issue is only that some studies could not be displayed.

## User impact
Users now see a clearer message that:
- explains some studies are unavailable right now
- reassures them they can still join the studies shown below
- avoids incorrectly blaming app versioning

## Root cause
The UI banner was tied to extraction failures for some study records, but the localization text assumed those failures always meant an outdated app.

## Validation
- traced the banner to `study_selection_hidden_studies`
- updated the English and German localization sources plus generated localization strings
- did not run automated tests for this copy-only change


johannesvedder (commented):
suggestion (non-blocking): keep outdated-version guidance in fallback copy

Serialization failures often happen when the app is too old for newer study data. The neutral copy is clearer, but dropping the update hint removes the most useful recovery action. Since #846 tracks a proper `upgrader` implementation, I would keep this banner as a fallback that mentions both possibilities:

```json
"study_selection_hidden_studies": "Some studies could not be displayed. This can happen when your app version is outdated. Please update the app to see all available studies, or join one of the studies shown below."
```

```json
"study_selection_hidden_studies": "Einige Studien konnten nicht angezeigt werden. Das kann passieren, wenn Ihre App-Version veraltet ist. Bitte aktualisieren Sie die App, um alle verfügbaren Studien zu sehen, oder nehmen Sie an einer der unten angezeigten Studien teil."
```

Follow-up for proactive update prompts: #846.

johannesvedder (changes requested):
suggestion(blocking): see comment above

mohiuddinshahrukh:
Sorry, didn't see this one, changes made.

johannesvedder (approved):
🚀

PR #819, opened 2026-04-22
fix(measurements): persist new survey questions

mohiuddinshahrukh:
## What changed
Fixes the measurements survey editor so newly created questions are saved correctly and duplicated questions no longer fail.

## Why
Creating a question in a brand-new measurement could appear to succeed, but the question would not remain visible because the route stayed on `/measurements/new` and reopened a fresh staged survey after save. Duplicating a fresh question could also fail because `responseOptionsValidity` was read before it had been initialized.

## User impact
Users can now:
- save a question in a newly created measurement survey and see it remain in the list
- duplicate a newly created question without triggering a runtime error
- continue editing on the saved measurement route instead of being reset to an empty staged survey

## Root cause
There were two issues in the measurements flow:
- question form duplication assumed `responseOptionsValidity` had always been initialized
- question changes in measurement surveys needed to propagate up to the parent survey save, and new measurements needed to redirect from the temporary `new` route to the real measurement id after the first save

## Validation
- reproduced the failing create-and-duplicate flow from the measurements page
- fixed the runtime crash in question duplication
- wired question save propagation for measurement surveys
- redirected newly saved measurements to their persisted route
- did not run automated tests

### DIFF
--- a/designer_v2/lib/features/design/measurements/measurements_form_controller.dart
+++ b/designer_v2/lib/features/design/measurements/measurements_form_controller.dart
@@ -194,6 +194,8 @@ class MeasurementsFormViewModel extends FormViewModel<MeasurementsFormData>
     MeasurementSurveyFormViewModel formViewModel,
     FormMode prevFormMode,
   ) async {
+    final isNewMeasurement = prevFormMode == FormMode.create;
+
     if (prevFormMode == FormMode.create) {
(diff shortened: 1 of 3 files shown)


johannesvedder (approved):
praise: Manual validation passed: creating a new measurement survey, adding and saving a question, URL redirect to persisted measurement route, duplicating the question, and console stayed error-free.

PR #820, opened 2026-04-25
fix(db): add ON DELETE CASCADE to study FK constraints

ibrahimozkn:
## Problem
Deleting a study from the backend threw a PostgrestException (23503 FK violation) when dependent records existed in `study_invite`, `study_subject`, `repo`, or `study_fitbit_credentials`. These four FK constraints referenced `study(id)` without `ON DELETE CASCADE`, blocking the delete entirely.

## Changes
- Added migration `database/migration/20260425_study_cascade_delete.sql` — drops and re-adds all four FK constraints with `ON DELETE CASCADE`
- Updated `database/studyu-schema.sql` to reflect the new constraint definitions

## Testing
- [ ] Run migration against Supabase instance and verify it applies cleanly
- [ ] Delete a study that has existing invites, subjects, repo entry, and Fitbit credentials — confirm it succeeds without FK error
- [ ] Verify dependent records are removed from all four tables after deletion
- [ ] No regressions in study creation or subject enrollment flows

### DIFF
--- a/database/migration/20260425_study_cascade_delete.sql
+++ b/database/migration/20260425_study_cascade_delete.sql
@@ -1,0 +1,24 @@
+--
+-- Add ON DELETE CASCADE to all FK constraints referencing study(id)
+-- so that deleting a study automatically removes all dependent records.
+--
+
+ALTER TABLE public.repo
+  DROP CONSTRAINT "repo_studyId_fkey",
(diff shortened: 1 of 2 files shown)


mohiuddinshahrukh (approved):
Everything looks good here, tested it locally and works fine 👍🏼

Issue #737, opened 2025-09-05
fix: Improve conditional questions


johannesvedder:
Fix the rest of the todos here: https://github.com/hpi-studyu/studyu/pull/730

- [ ] Try to break stuff, e.g. delete a question that is part of a conditional and check if the condition on the question gets removed or if there is an error
- [ ] Test eligibility functionality in the app
- [ ] duplication of new survey item does only work after saving the questionnaire
- [ ] On conditions launch check if conditions are valid (targets and values exist and make sense). Otherwise delete the affected condition -> See https://github.com/hpi-studyu/studyu/pull/732

ibrahimozkn:
In launched studies, the UI currently allows deleting questions. However:
	•	There is no Save button, so the deletion is not persisted in the backend.
	•	If the deleted question had conditions (visibility logic) on another question, it remains on the visibility condition list as "Question" item.

<img width="1490" height="1912" alt="Image" src="https://github.com/user-attachments/assets/4456ad4f-b6f8-4cfb-9c64-157c0929ee55" />

PR #829, opened 2026-05-04
fix(designer): duplicate unsaved survey questions

mohiuddinshahrukh:
## What changed

- Save the duplicated form view model instead of re-saving the original when using the duplicate action.
- Initialize question response option validity with an empty map so newly created, unsaved questions can be copied before they have been loaded from the domain model.
- Add focused coverage for the duplicate action and copying newly created question data.

## Why

Part of #737. Duplicating a newly added survey question could fail before the questionnaire itself was saved. Saved questions worked because their response option validity was initialized during domain-model loading; newly created questions did not go through that path.

## Testing

- `flutter test test\features\forms\form_view_model_collection_actions_test.dart`
- Manual Designer frontend test: added a new unsaved survey question, duplicated it before saving the parent questionnaire, and confirmed the duplicate appears correctly.

### DIFF
--- a/designer_v2/lib/features/design/shared/questionnaire/question/question_form_data.dart
+++ b/designer_v2/lib/features/design/shared/questionnaire/question/question_form_data.dart
@@ -94,7 +94,7 @@ abstract class QuestionFormData implements IFormData {
   final QuestionConditional? conditional;
 
   /// Mapping from response option => qualifying/disqualifying
-  late final Map<dynamic, bool> responseOptionsValidity;
+  Map<dynamic, bool> responseOptionsValidity = {};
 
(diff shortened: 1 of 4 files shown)


johannesvedder (approved):
Praise: Clean, minimal fix for part of #737 

- formViewModel.save() to duplicateFormViewModel.save(): exactly the right one-line fix. The original code persisted the wrong VM, making duplicates of unsaved items a no-op.
- late final to Map<dynamic, bool> = initializes responseOptionsValidity: removes a latent crash path for unsaved questions without adding ceremony.
- Test coverage directly exercises both the fix and the edge case (copy before initialization). No mock overkill, pure test doubles with an observable saveCount.

Manual and automated tests pass. Looks correct.

PR #831, opened 2026-05-11
fix(db): scope invite codes to studies

mohiuddinshahrukh:
## Description

Invite-code validation could rely on `invite_code` alone, so enrollment could reference an invite code without the database also enforcing that the code belongs to the same study.

## Changes

- Scope app invite lookup by both `code` and `study_id`
- Add composite unique constraint on `study_invite (study_id, code)`
- Replace `study_subject.invite_code` foreign key with composite `(study_id, invite_code)` foreign key
- Add `public.is_invite_code_for_study(uuid, text)` as a `SECURITY DEFINER` helper
- Update restrictive insert policy on `study_subject` to validate `(study_id, invite_code)`
- Revoke function execution from `public` and `anon`
- Mirror DB changes in `database/studyu-schema.sql`

## Testing

- `flutter analyze` in `app`
- `flutter analyze` in `designer_v2`
- `git diff --check`
- Frontend invite-code flow:
  - invalid code rejected
  - valid code accepted
- Applied migration to local Supabase
- Transaction-only DB test as `postgres`:
  - valid `(study_id, invite_code)` accepted
  - mismatched pair rejected
  - rolled back
- Authenticated/RLS-level DB test:
  - valid pair accepted under `authenticated`
  - mismatched pair rejected by policy

### DIFF
--- a/app/lib/screens/study/onboarding/study_selection.dart
+++ b/app/lib/screens/study/onboarding/study_selection.dart
@@ -300,6 +300,7 @@ class _InviteCodeDialogState extends State<InviteCodeDialog> {
                 .from('study_invite')
                 .select('preselected_intervention_ids')
                 .eq('code', _controller.text)
+                .eq('study_id', study.id)
                 .maybeSingle();
             if (!context.mounted) return;
             if (inviteResult != null &&
--- a/database/migration/20260506_scope_invite_codes_to_study.sql
+++ b/database/migration/20260506_scope_invite_codes_to_study.sql
@@ -1,0 +1,61 @@
+BEGIN;
+
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1
+    FROM pg_constraint
+    WHERE conname = 'study_invite_study_id_code_unique'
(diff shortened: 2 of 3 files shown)


johannesvedder:
I think this changes the invite-code model in a way that does not match our enrollment flow. Users can enter an invite code before they know or can access the study; the invite code itself is used to determine which study they can join. Therefore invite codes need to be globally unique across the platform, not only unique per (study_id, code). A composite unique constraint is useful for the FK/RLS check, but it should not replace a global UNIQUE (code) constraint. Otherwise the same code could exist for multiple studies and code-first enrollment becomes ambiguous.

johannesvedder:
We could of course combine uniqueness of invite code and study id but this would enforce changing the enrollment model and combine it with e.g the #781 PR so participants Don't have to input study id and the invite code. But this would require a new issue.

johannesvedder (commented):
I appreciate the idea and definitely see that this needs to be improved, but would close this for now and create an issue for the new task.

mohiuddinshahrukh:
Makes sense!
Thank you for reviewing!

PR #840, opened 2026-05-28
STUDYU-65 feat: dashboard pagination

ibrahimozkn:
This pull request introduces significant improvements to the dashboard's architecture, state management, and navigation indirection, as well as some smaller UI and test updates. The most notable change is the decoupling of the dashboard controller from the app router, which resolves test compilation issues and improves modularity. Additionally, dashboard state handling (including filter, sort, search, and pinning) is refactored for clarity and efficiency, with better persistence and debouncing. There are also minor UI and test improvements.

## Video
https://github.com/user-attachments/assets/e45dc5d2-c1a7-4d7c-ae9e-47075c38d396

## Summary by CodeRabbit

* **New Features**
  * Dashboard now supports paginated study loading, search, server-side filtering, pinned studies, and persisted sorting.
  * Added loading, retry, end-of-list, empty-state, and unsupported-filter messaging.
  * Study actions now refresh results and improve navigation to study editing.
  * Added a study preview layout and improved help-icon hover behavior.

* **Bug Fixes**
  * Study deletion and closing now wait for persistence and surface failures reliably.

* **Localization**
  * Added English and German translations for dashboard and study-list messaging.

* **Changes**
  * Filter match counts are no longer displayed.
  * Missed-days filtering is no longer supported.


mohiuddinshahrukh:
praise: Tested and everything looks good! 👍 
Edit: Upon further testing, I have come across some issues, see the comments that I make futher from here please.

mohiuddinshahrukh (dismissed):

mohiuddinshahrukh:
## Summary
I tested this locally in the browser with many studies in the DB.

- Before the fixes (that I had to make myself), pagination did **not** work reliably.
- The list often stopped around 25 items.
- More studies were not loaded while scrolling.
- Sometimes the “End of list” message did not appear.

After applying fixes, pagination now works:
- I can scroll and load later studies (for example Study 40+ and Study 60).
- “End of list” is shown when the final page is reached.

---

## Main Issue Found
The page was listening to the wrong scroll area.  
So `loadMore()` was not triggered when the actual studies list was scrolled.

---

## What Was Changed
1. Connected pagination to the **real right-side scroll container** (the one the user actually scrolls).
2. Removed an extra loading gate that could hide the table even when data existed.
3. Re-tested in browser + ran tests/analyze.

---

## Conventional Comments

- `issue:` I could reproduce this bug in browser. Scrolling stopped loading new studies after the first page.
- `suggestion:` Please keep the scroll listener attached to the real list container, not the page context.
- `suggestion (non-blocking):` Add one UI/integration test that scrolls and confirms `loadMore()` is called.

---

@ibrahimozkn

ibrahimozkn:
> ## Summary
> I tested this locally in the browser with many studies in the DB.
> 
> * Before the fixes (that I had to make myself), pagination did **not** work reliably.
> * The list often stopped around 25 items.
> * More studies were not loaded while scrolling.
> * Sometimes the “End of list” message did not appear.
> 
> After applying fixes, pagination now works:
> 
> * I can scroll and load later studies (for example Study 40+ and Study 60).
> * “End of list” is shown when the final page is reached.
> 
> ## Main Issue Found
> The page was listening to the wrong scroll area. So `loadMore()` was not triggered when the actual studies list was scrolled.
> 
> ## What Was Changed
> 1. Connected pagination to the **real right-side scroll container** (the one the user actually scrolls).
> 2. Removed an extra loading gate that could hide the table even when data existed.
> 3. Re-tested in browser + ran tests/analyze.
> 
> ## Conventional Comments
> * `issue:` I could reproduce this bug in browser. Scrolling stopped loading new studies after the first page.
> * `suggestion:` Please keep the scroll listener attached to the real list container, not the page context.
> * `suggestion (non-blocking):` Add one UI/integration test that scrolls and confirms `loadMore()` is called.
> 
> @ibrahimozkn

question: it worked when I tested. Could you please send a recording or retest?

mohiuddinshahrukh:
https://github.com/user-attachments/assets/10115e62-6a74-43bf-954c-7c2b08c80ee6

mohiuddinshahrukh (dismissed):
praise: After doing melos BS, the implemented feature is working!

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -234,30 +254,70 @@ class StudiesTable extends StatelessWidget {
...
+                              height: 22,
+                              child: CircularProgressIndicator(strokeWidth: 2),
+                            ),
+                          ),
+                        ),
+                        noMoreItemsIndicatorBuilder: (_) => Padding(
+                          padding: const EdgeInsets.symmetric(vertical: 16.0),
+                          child: Center(
+                            child: Text(
+                              "End of list".hardcoded,
```
johannesvedder: Needs localization. Also, the EoL message sounds a bit unpolished. Maybe replace it with something warmer and actionable, like:

“That’s all for now. Want to `[create a new study](link-to-new-study-page)`?”

German:

“Das war’s fürs Erste. `[Neue Studie erstellen](link-to-new-study-page)`.”

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -234,30 +254,70 @@ class StudiesTable extends StatelessWidget {
...
+                            ),
+                          ),
+                        ),
+                        newPageErrorIndicatorBuilder: (_) => Padding(
+                          padding: const EdgeInsets.symmetric(vertical: 16.0),
+                          child: Center(
+                            child: TextButton.icon(
+                              onPressed: onRetry,
+                              icon: const Icon(Icons.refresh),
+                              label: Text("Retry".hardcoded),
```
johannesvedder: Needs localization as well

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -312,3 +372,75 @@ class StudiesTable extends StatelessWidget {
...
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            Icon(
+              Icons.filter_alt_off_outlined,
+              color: Theme.of(context).hintColor,
+            ),
+            const SizedBox(height: 8),
+            Text(
+              "This filter cannot be applied server-side and is not currently supported."
+                  .hardcoded,
```
johannesvedder: Weird error message. Why are some filters not supported?

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -312,3 +372,75 @@ class StudiesTable extends StatelessWidget {
...
+              textAlign: TextAlign.center,
+              style: Theme.of(context).textTheme.bodyMedium,
+            ),
+            if (onRetry != null) ...[
+              const SizedBox(height: 8),
+              TextButton(
+                onPressed: onRetry,
+                child: Text("Reset filter".hardcoded),
```
johannesvedder: .

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -312,3 +372,75 @@ class StudiesTable extends StatelessWidget {
...
+    return Padding(
+      padding: const EdgeInsets.symmetric(vertical: 24.0),
+      child: Center(
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const Icon(Icons.error_outline),
+            const SizedBox(height: 8),
+            Text(
+              "Failed to load studies.".hardcoded,
```
johannesvedder: .

designer_v2/lib/features/dashboard/studies_table.dart
```diff
@@ -312,3 +372,75 @@ class StudiesTable extends StatelessWidget {
...
+              textAlign: TextAlign.center,
+              style: Theme.of(context).textTheme.bodyMedium,
+            ),
+            if (onRetry != null) ...[
+              const SizedBox(height: 8),
+              TextButton.icon(
+                onPressed: onRetry,
+                icon: const Icon(Icons.refresh),
+                label: Text("Retry".hardcoded),
```
johannesvedder: .

designer_v2/lib/features/dashboard/dashboard_state.dart
```diff
@@ -12,23 +11,59 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 
 class DashboardState extends Equatable {
   static const defaultFilter = StudiesFilter.owned;
+  static const pageSize = 25;
```
johannesvedder: Maybe cut this down to 15-20 to reduce initial loading time even further? At least on my mac only 8 studies are shown at the same time. Would need to test this on different resolutions. What happens if pageSize < number of studies that fit on the screen initially?

johannesvedder (changes requested):
Thanks! Works nicely. Some things I found:

- perf: Scrolling the whole studies page feels sluggish even though all studies have loaded once. On the recorded videos it's much more severe than without the recording, but it still is a bit annoying. Seems like a performance problem of the library or the implementation?

https://github.com/user-attachments/assets/239d7a9e-14a1-4424-93f1-9e593ec207eb

- bug: Infinite loading screen shows up for filter "Public Studies". Maybe the no studies found error doesn't show up?

Image

- QoL: On loading the study registry, error shows up briefly

https://github.com/user-attachments/assets/58e0c996-5df1-4215-8dbf-33b019294d5d

johannesvedder:
Maybe add to PR template and/or AGENTS.md that no hardcoded strings exist and everything is localized?

johannesvedder:
Also should we have a label to show how many studies are there in the list? How can this be implemented without causing too much noise?

johannesvedder:
Todo:

- [x] Verify that study pinning works with these changes (#817):

> We must verify infinite scrolling doesn't interfere with pinned studies, e.g. pinned studies should always load first in the study list so they need to be prioritized on server side. Filtering + sorting is now on server side so pinning should be too I guess? Also check if there are other edge cases we need to take care of e.g. what happens if there are more pinned studies than the initially loaded number of studies and so on.

mohiuddinshahrukh:
@johannesvedder 
@ibrahimozkn 

issue: Regarding the pinning: It does not work as it should right now.

issue: One more thing that came to my observation: The pinned studies cannot be searched, I just see a spinner for an infinite amount of time without any error or message.

issue: The pinning icon is also still broken - Even if I refresh the page after pinnning a row, I cannot see the pinned icon.

If a video is required, please let me know.

mohiuddinshahrukh:
https://github.com/user-attachments/assets/1307991e-e2d8-400b-a03e-3342988eb6c4

To test this PR: 

- [ ] Open the studies, 
- [ ] Scroll
- [ ] Make sure lag is not unbearable
- [ ] Check that the behavior is working with respect to responsive layout
- [ ] Finally check that study pinning is working

For more reference you can look at the video also.

johannesvedder:
@mohiuddinshahrukh already looks good! As we discussed I would phrase it as "x studies" where x is the total number of studies of the current page (my studies or studies registry). If a filter is active the label can be changed to "y of x studies".

Also could you fix all the other tasks from my earlier review please before I do a final review. Thanks

designer_v2/lib/features/dashboard/dashboard_state.dart
mohiuddinshahrukh: Cut down to 20

ibrahimozkn:
Addressed all the blocking issues from @johannesvedder's review. Here's what changed:

**Pagination**
- `hasMore` wasn't being cleared on failed load-more requests, so every scroll after an error kept retrying. Fixed.
- Added `id` as a secondary sort key. Without it, tied rows on `created_at` or `status` could come back in different order across pages, causing duplicates or gaps.

**Mutations**
- Delete, duplicate, and close now trigger a full state reload after they complete. Deleted studies were staying in the list, duplicates weren't showing up.
- Delete errors are now surfaced to the user via `loadError`. They were silently dropped before.

**Pinned studies**
- Replaced the manual list manipulation in `pinStudy`/`pinOffStudy` with a full reload. The old approach modified `loadedStudies` directly, which shifted the offset for the next page fetch and caused pinned rows to bypass search and filter.

**UI**
- Search field was collapsing — `Flexible` inside a `MainAxisSize.min` row gets no flex space. Replaced with `SizedBox(width: searchMaxWidth)`.
- Count label was flashing "1 of 11" on search input because `filteredPinnedStudyCount` settled before `totalCount` came back from the server. Added an `isLoadingInitial` guard so the filtered format only shows once the fetch is done.
- Public and Shared Studies tabs were returning `SizedBox.shrink()` on empty results. Both now show a proper empty state with localized strings.

**Filters**
- `FilterOperator.equals` on string columns was using `eq` (case-sensitive). Switched to `ilike`. Also escaping `%` and `_` in wildcard patterns now.

**Localization**
- `"Entwurf"` was hardcoded for column width math. Replaced with `tr.study_status_draft`.
- "Add filter" / "Remove filter" tooltips in `filter_item.dart` were `.hardcoded`. Moved to arb files with DE translations.

**Test**
- Updated the assertion that expected `fetchPage` called once — `_resetAndReload` fires both a data call and a count call now, so the correct expectation is 2. Also rewrote the `pinStudy` test to match the reload-based behavior.

johannesvedder:
@mohiuddinshahrukh could you review first and then let me do a final review before @ibrahimozkn merges please

mohiuddinshahrukh:
issue:
image
image

browser_client.dart:83 
 GET http://localhost:8082/rest/v1/study?select=id%2Ctitle%2Cdescription%2Cuser_…22%29&order=created_at.desc.nullslast%2Cid.asc.nullslast&offset=0&limit=20 404 (Not Found)

Supabase Exception encountered
js_primitives.dart:28 42883
js_primitives.dart:28 Not Found
js_primitives.dart:28 operator does not exist: study_status ~~* unknown

issue: Making duplicate of study doesn't show it till page is refreshed. 
issue: Study deletion doesn't work reliably
Issue: Page redirection, when I delete a study on "Shared with me tab" - I get redirected to "My Studies"
issue: Copy as draft working unreliably

comment[non-blocking]: Regarding the issues of unreliabilty I think there is an update state not working correctly therefore we get stale states - this should be a straight forward page state update fix for all these problems.

johannesvedder (approved):
praise: dashboard pagination follow-ups are ready

The study count no longer flashes 0 during reload, and pinning a loaded study now updates locally while unpinning and unsafe cases retain the safe reload path. The focused Chrome tests and targeted analysis pass.

Approved from my side. @mohiuddinshahrukh, please do a final review before merge.

mohiuddinshahrukh:
https://github.com/user-attachments/assets/9692ddbb-90f8-457b-991a-47da3d455d13

This is still an issue. 
Same goes for duplication - these are not working reliably.

I can also fix this @johannesvedder

johannesvedder:
> Screen.Recording.2026-07-24.at.12.40.36.mov 
> This is still an issue. Same goes for duplication - these are not working reliably.
> 
> I can also fix this @johannesvedder

I see. Yeah sure would be great if you could take over this. Just ping me for review when you are done. Thanks!

johannesvedder (changes requested):
praise: Visual testing was working. Deleting a study + filtering actions worked well.
Agentic code review result:

```
 issue (blocking): make legacy study deletion atomic

 designer_v2/lib/repositories/api_client.dart:379-405 deletes dependent records through
 separate committed requests before retrying the study deletion. If an intermediate
 request or the final delete fails, the study remains but some participant or repository
 data has already been permanently deleted. Remove the client-side fallback and require ON
 DELETE CASCADE, or perform the cleanup and deletion in one transactional database
 function.

 issue (blocking): filter pinned studies before rendering them

 designer_v2/lib/features/dashboard/dashboard_state.dart:118 concatenates every pinned
 study with the paginated results, although the derived counts filter pinned studies by
 page, search, and advanced filter. Since the server query excludes pinned IDs, unrelated
 pinned studies appear under searches and Public/Shared tabs while the matching paginated
 row is omitted. Apply _matchesPagePreset, _matchesSearchQuery, and _matchesActiveFilter
 before adding pinned studies to displayedStudies.

issue (blocking): include collaborator metadata in dashboard queries

 designer_v2/lib/repositories/api_client.dart:132-146 omits collaborator_emails from
 studyDisplayColumns, used by paginated and pinned queries at lines 233 and 288. Shared
 studies therefore deserialize with an empty collaborator list, causing collaborator
 actions such as edit and copy to disappear and client-side preset checks to fail. Add
 collaborator_emails to the projection.

 issue (blocking): preserve unsupported-filter exceptions across the API layer

 designer_v2/lib/repositories/api_client.dart:243-275 lets buildPostgrestFilterExpression
 throw UnsupportedFilterException, then converts it into a generic APIException in the
 final catch. The controller’s typed catch is therefore unreachable for real API calls,
 and the dedicated unsupported-filter state is not shown. Rethrow
 UnsupportedFilterException before the generic catch.

 issue (blocking): do not silently discard auxiliary-request failures

 designer_v2/lib/features/dashboard/dashboard_controller.dart:119-121,263-265 suppresses
 pinned-study and page-count errors. When pinned loading fails, the page query still
 excludes every configured pinned ID, making those studies disappear entirely. A failed
 count request can likewise show zero while rows are visible. Surface a retryable error
 and either preserve previous values or fetch the page without pinned exclusions when
 pinned loading fails.

 issue (non-blocking): make the unsupported-filter clear button clear the filter

 designer_v2/lib/features/dashboard/studies_table.dart:430-432 labels the action “Clear”
 but calls onRetry, which is wired to DashboardController.retry at
 dashboard_page.dart:633. Clicking it reruns the same unsupported filter and returns to
 the same notice. Wire the button to updateFilter(FilterGroup()); keep retry for transient
 failures.

 issue (non-blocking): retain drawer scrolling on short layouts

 designer_v2/lib/features/app_drawer.dart:253 adds NeverScrollableScrollPhysics to the
 navigation list. On short viewports or increased browser/text zoom, entries can become
 clipped and unreachable. Keep the drawer scrollable, for example with
 ClampingScrollPhysics.
```

@mohiuddinshahrukh

johannesvedder:
@ibrahimozkn or @mohiuddinshahrukh you may review

PR #844, opened 2026-06-02
refactor(app): simplify questionnaire flow

johannesvedder:
## Description
Updates #844.

Refactors questionnaire answer handling into a controller-driven flow with one inline CTA and adds review handling for restored cached answers.

### Questionnaire flow
- Adds `QuestionnaireController` as the source of truth for committed answers, free-text drafts, visibility, payload building, answer metadata, and CTA mode.
- Removes the parent duplicate Complete button in `questionnaire_task_widget.dart` and per-field free-text Submit actions.
- Uses one questionnaire CTA:
  - orange `Continue` commits drafts and reveals the next visible question
  - green `Complete` submits the visible payload
- Requires the user to press `Complete`; answering the last question no longer auto-submits.
- Keeps hidden answers cached for the current session, restores supported UI answers/drafts, and excludes hidden answers from the submitted payload.
- Prevents hidden cached answers from driving downstream visibility unless a hidden default answer is part of the question definition.

### Restored-answer review
- Adds per-answer metadata in `QuestionnaireState.answerMetadata`:
  - `restoredFromCache`
  - `needsReview`
  - `cacheContext`
- Marks restored lower answers as needing review when an earlier answer changes.
- Stores review state in form state, not widget-only state.
- Shows `Please review this restored answer.` beside visible answers that need review.
- Adds `Mark as reviewed` to clear the flag.
- Blocks `Complete` while a visible answer still needs review.
- Hidden answers that need review do not block submit until they become visible again.

### Review behavior examples
- Medication branch:
  - Q1: Did you take medication?
  - Q2: Which medications? shown only if Q1 = yes
  - Q3: What did you eat today?
  - With the no-metadata rule, changing Q1 marks later restored answers for review. This favors stale-data prevention over trying to infer independent questions.
- Meal report:
  - Q1: Which meal are you reporting? Breakfast/Lunch/Dinner
  - Q2: What did you eat?
  - Changing Q1 from Breakfast to Dinner keeps the cached Q2 text but marks it for review before submit.
- Pain score:
  - Q1: Rate your pain 0-10
  - Q2: Have you taken painkillers? shown only if Q1 > 5
  - Q1 6 → Q2 yes → Q1 2 hides Q2 → Q1 6 restores Q2 without review because the context returned to the original value.
  - Q1 6 → Q2 yes → Q1 7 keeps Q2 visible and marks it for review.
  - Q1 7 → Q1 6 clears review because the context returned to the answer's stored context.

### Implementation notes
- Without question-level dependency metadata, the app cannot safely tell whether a later question depends semantically on an earlier question.
- The controller therefore treats all earlier answers as the cache context for a question.
- Any answered/drafted visible question below a changed question can require review.
- This makes the behavior conservative: users may review an independent lower answer, but they cannot submit stale lower answers unnoticed.
- A future schema-level dependency field would allow narrower review behavior without guessing from order or prompt text.

### Other fixes kept in this PR
- CTA styling: readable buttons, centered, not full-width, inline after the last question.
- Validation only fires on text change, not first focus.
- Removed free-text input formatters for numeric/alphanumeric fields; users now see validation errors instead of silent keypress rejection.
- Restored free-text drafts that mirror committed answers no longer force `Continue`.
- Added placeholder hint text on empty free-text fields.
- Added `isSubmitting`; the CTA shows a spinner while the parent processes submission.

## Testing Steps
- [x] `fvm flutter test test/widgets/questionnaire/questionnaire_widget_test.dart test/widgets/questionnaire/questionnaire_controller_test.dart`
- [x] Dart MCP analyze for changed questionnaire Dart files
- [x] `fvm exec melos run qualitycheck`

### PR Checklist
- [x] `fvm exec melos run qualitycheck` passes
- [ ] Screenshot or video of the changes attached
- [x] Description links related issues


mohiuddinshahrukh:
praise: Tested the changes made. 
Everything looking good and as expected!

mohiuddinshahrukh (approved):

mohiuddinshahrukh (approved):

PR #871, opened 2026-07-01
STUDYU-61 fix(designer): support free-text length conditions

mohiuddinshahrukh:
## Description

Adds free-text visibility conditionals based on character length and fixes related questionnaire resubmission behavior.

Changes:
- adds free-text comparators for `>`, `<`, `>=`, `<=` using answer character length
- evaluates free-text length conditionals in core expression logic
- restricts length-comparator condition values to numeric thresholds in designer
- prevents stale non-numeric comparator values from being saved when switching to length comparators
- localizes live preview labels for free-text length comparators
- restores inline `Done` for edited submitted free-text answers so users can resubmit after changing an answer

## Testing Steps

- [x] Run `fvm dart run melos run qualitycheck`

1. In Designer, create a free-text question `Q1`.
2. Create a later question `Q2`.
3. Add visibility logic to `Q2` using `Q1` with comparator `>` and value `0`.
4. Save and verify preview text shows a length-based condition.
5. In app preview, leave `Q1` empty and verify `Q2` stays hidden.
6. Enter one or more characters in `Q1` and verify `Q2` appears.
7. Change a submitted free-text answer and verify inline `Done` appears again so the edited answer can be resubmitted.
8. In Designer, enter a text comparator value like `asdasd`, switch comparator to `>=`, and verify stale text is cleared and not saved.

https://github.com/user-attachments/assets/3e280fc6-0457-428d-b450-e5003002b9f4

https://github.com/user-attachments/assets/1cb4b4f1-b76e-47d9-beda-e12b8cd55fc4

image
image


johannesvedder (changes requested):
issue (blocking): use type-specific comparator labels for free-text conditions

`<`, `>`, `<=`, and `>=` currently evaluate `targetValue.length`, not the numeric value of the answer. That makes a numeric-looking condition on a free-text question misleading: authors will read `Q1 > 5` as "answer value greater than 5", while the runtime reads it as "answer length greater than 5".

Can we split the dropdown by free-text subtype?

- if the referenced free-text question is `FreeTextQuestionType.numeric`, show numeric comparators: `<`, `>`, `<=`, `>=`, `=`, `!=`; evaluate with `NumericExpression`/numeric parsing semantics
- if the referenced free-text question is anything else, hide symbolic numeric comparators and show text/length labels: contains, does not contain, shorter than, shorter than or equal to, longer than, longer than or equal to, same length as, different length as

This keeps non-CS authors from needing to know that symbolic operators on strings mean character length here.

issue (blocking): new widget test leaves a pending timer

`app/test/widgets/questionnaire/questions/free_text_question_widget_test.dart:152` fails locally because the new test exits with `_ensureTextFieldVisible`'s 500 ms timer still pending. The previous nearby test drains it with `await tester.pump(const Duration(milliseconds: 600));`; this test needs the same cleanup or a safer timer strategy.

Verification:
`fvm flutter test app/test/widgets/questionnaire/questionnaire_controller_test.dart app/test/widgets/questionnaire/questions/free_text_question_widget_test.dart`

Result: `Done button appears after editing submitted non-last free text` fails with `A Timer is still pending even after the widget tree was disposed.`

johannesvedder:
@mohiuddinshahrukh is this ready for review again?

app/lib/widgets/questionnaire/questionnaire_controller.dart
```diff
@@ -233,7 +235,20 @@ class QuestionnaireController extends ChangeNotifier {
...
-    return conditional.condition.evaluate(evaluationState) == true;
+    final result = conditional.condition.evaluate(evaluationState);
+    debugPrint(
+      '$_questionnaireDebugLogPrefix visibility '
+      'question=${question.id} result=$result '
+      'condition=${conditional.condition.toJson()} '
+      'answers=${_debugAnswerSnapshot(evaluationState)}',
+    );
+    return result == true;
+  }
```
johannesvedder: Can you remove all debugPrints please?

johannesvedder: - eligibility_screen.dart:63-65 logs eligibility answers
 - questionnaire_controller.dart:239-243 logs visibility answers
 - questionnaire_controller.dart:319-321 logs submitted answers
 - questionnaire_controller.dart:368-370 logs free-text drafts

johannesvedder: you can wrap in assert(() { ...; return true; }()); and avoid raw responses.

app/lib/screens/study/onboarding/eligibility_screen.dart
```diff
@@ -58,8 +60,25 @@ class _EligibilityScreenState extends State<EligibilityScreen> {
...
+      element,
+    ) {
+      final result = element.condition.evaluate(qs);
+      debugPrint(
+        '$_eligibilityDebugLogPrefix criterion '
+        'id=${element.id} result=$result '
+        'condition=${element.condition.toJson()}',
+      );
+      return result == false;
+    });
```
johannesvedder, line 66: Could you check this please:

issue (blocking): app/lib/screens/study/onboarding/eligibility_screen.dart treats unevaluable eligibility criteria as passing

 _evaluateResponse now fails only when criterion.condition.evaluate(qs) == false (eligibility_screen.dart:115-123). Before, criterion.isSatisfied(qs) required evaluate(qs) == true (EligibilityCriterion.isSatisfied).
 So null now becomes eligible. Missing answers, malformed values, or type mismatches can let participants pass eligibility instead of failing/blocking.

 Fix: keep final eligibility on criterion.isSatisfied(qs) semantics, or fail on result != true.

designer_v2/lib/features/design/shared/questionnaire/question/question_conditional_form_view.dart
```diff
@@ -377,6 +405,12 @@ class ConditionalQuestionFormView extends FormConsumerWidget {
       case FreeTextQuestion.questionType:
         return ReactiveTextField<dynamic>(
           formControl: conditionVm.valueControl,
+          keyboardType: conditionVm.selectedComparatorUsesNumericThreshold
+              ? TextInputType.number
+              : TextInputType.text,
+          inputFormatters: conditionVm.selectedComparatorUsesNumericThreshold
+              ? [FilteringTextInputFormatter.digitsOnly]
+              : null,
           decoration: InputDecoration(
```
johannesvedder: issue (blocking): designer numeric free-text condition UI blocks valid negative thresholds

 FreeTextQuestionType.numeric accepts negative integers (free_text_question.dart:88-90, regex ^-?[0-9]+$).
 But conditional threshold input for numeric free-text uses FilteringTextInputFormatter.digitsOnly (question_conditional_form_view.dart:411-413), so designers cannot enter conditions like “numeric answer < -1” even though runtime
 NumericExpression and num.tryParse support it.

 Fix: use the same signed numeric formatter as scale where appropriate, or at least allow - for numeric free-text comparator values.

johannesvedder (changes requested):
Visual testing looked nice for questionnaire. I still need to test eligibility. Review agent found some blocking issues I pasted to the code lines.

When I was testing with a testing questionnaire, the questionnaire display logic was messed up. I think my last refactor is still not finished, we need to do another revision on this, but this is out of scope for this PR.

app/lib/widgets/questionnaire/questionnaire_controller.dart
johannesvedder: > * eligibility_screen.dart:63-65 logs eligibility answers
> 
>     * questionnaire_controller.dart:239-243 logs visibility answers
> 
>     * questionnaire_controller.dart:319-321 logs submitted answers
> 
>     * questionnaire_controller.dart:368-370 logs free-text drafts

Sorry I meant to to remove all the debugPrint's stated here not all debugPrint's of the whole app. ^^ @mohiuddinshahrukh

johannesvedder:
issue (blocking): numeric threshold input still uses an unsigned software keyboard

The formatter in `designer_v2/lib/features/design/shared/questionnaire/question/question_conditional_form_view.dart` accepts negative thresholds, but the field still uses `TextInputType.number`. Mobile keyboards may therefore omit the minus key.

Use `TextInputType.numberWithOptions(signed: true)` when `selectedQuestionAllowsSignedNumericThreshold` is true.

johannesvedder (changes requested):
praise: Functionality works as intended

Eligibility for free text fields is not yet implemented, so there is nothing further to test.
Please just fix the rest of the issues.

johannesvedder:
I found two more issues and just quickly resolved them myself. PR looks fine to me otherwise!

```
 issue (blocking): preserve existing numeric free-text conditions when loading the Designer

 New numeric free-text conditions correctly use NumericComparator and NumericExpression. However, existing conditions
 may contain TextComparator.equal, notEqual, contains, or doesNotContain.

 _normalizeLegacyFreeTextComparator at
 designer_v2/lib/features/design/shared/questionnaire/question/question_conditional_row_form_controller.dart:461-476
 only converts the new length comparators. The numeric expression builder at lines 401-407 rejects the remaining text
 comparators and returns null, so opening and saving an existing questionnaire can silently remove its condition.

 Map legacy equal and notEqual to their numeric equivalents and add a load/round-trip regression test. Define a
 non-lossy compatibility policy for legacy contains and doesNotContain.

 note (non-blocking): restore direct coverage for unevaluable eligibility criteria

 The implementation correctly uses !criterion.isSatisfied(qs), but the current eligibility test no longer directly
 exercises an evaluation result of null.
```

johannesvedder (approved):

PR #889, opened 2026-07-16
STUDYU-55 chore: merge main into dev after invite hotfix

johannesvedder:
## Summary

- merge current `main` into `dev`, including #888
- preserve `dev`'s `Study.fetchByInviteCode` and GoRouter architecture
- snapshot the submitted invite code before the async lookup so the loaded study and stored invite code cannot diverge
- wrap the descriptions in all four package pubspecs using YAML folded scalars

## Conflict resolution

`dev` already performs both invite-related database requests inside `Study.fetchByInviteCode` before dismissing the dialog. The specific #888 race therefore does not occur on `dev`: there is no await between `context.pop()` and the synchronous state update/navigation call.

The merge keeps that implementation instead of copying `main`'s direct Supabase queries and Navigator routing.

## Designer dependency review

#888 pinned `custom_lint: 0.8.1` and `riverpod_lint: 3.0.3` for the older `main` dependency graph. `dev` deliberately removed those dependencies while upgrading Flutter in `1daaced1c`, and neither package exists in the current workspace lockfile. The merge therefore keeps `dev`'s newer dependency set instead of reintroducing the pins.

The package descriptions in `app`, `core`, `designer_v2`, and `flutter_common` now use `>-` folded scalars. Their parsed values are unchanged. The short root workspace description remains on one line.

## Verification

- `fvm flutter pub get`
- `fvm flutter analyze`
- `fvm flutter analyze app/lib/screens/study/onboarding/study_selection.dart`
- `cd app && fvm flutter test test/widget_test.dart`
- parsed all changed pubspecs with Ruby YAML
- `git diff --check`

All passed locally.


mohiuddinshahrukh (approved):
praise: Tested the invite code flow specifically - everything looks good.

PR #892, opened 2026-07-16
STUDYU-54 fix(designer): handle study schedule number input

johannesvedder:
## Description
Changing the study schedule's numeric fields could throw two web runtime errors:

- Values above the configured range were clamped without updating the text selection, leaving the cursor outside the replacement text.
- Clearing a field passed an empty string to `int.parse`.

This change places the cursor at the end of clamped values and ignores transient empty edits until the user enters another number. A widget test covers both interactions for the number-of-cycles field.

## Testing Steps
1. Open a draft study and go to **Design → Interventions**.
2. Append a digit to **Number of cycles** so the value exceeds its maximum. Confirm it clamps to `9` without an assertion.
3. Clear the field, then enter another valid number. Confirm no `FormatException` occurs.

Automated checks:
- `rtk fvm flutter test test/utils/input_formatter_test.dart test/features/design/interventions/study_schedule_form_view_test.dart`
- `scripts/pre-commit-check`

### PR Checklist
- [x] Formatting and analyzer pass via `scripts/pre-commit-check`
- [ ] Screenshot or video attached

### DIFF
--- a/designer_v2/lib/features/design/interventions/study_schedule_form_view.dart
+++ b/designer_v2/lib/features/design/interventions/study_schedule_form_view.dart
@@ -160,38 +160,26 @@ class _StudyScheduleFormViewState extends State<StudyScheduleFormView> {
                   tr.form_field_crossover_schedule_num_cycles_tooltip,
               input: Row(
                 children: [
-                  Container(
-                    constraints: const BoxConstraints(maxWidth: 70),
-                    child: TextField(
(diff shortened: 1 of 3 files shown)


mohiuddinshahrukh (approved):
suggestion [non-blocking]: why not make this a dropdown from 1-9?
Much easier to use it then rather than entering numbers manually?

Image

johannesvedder:
Screenshot 2026-07-22 at 12 33 15
Changed design

PR #897, opened 2026-07-16
STUDYU-52 feat(app): support required choice selections

johannesvedder:
## Description
Multi-select choice questions currently allow participants to confirm an empty selection. Researchers cannot distinguish an intentionally empty response from a missed question, and the designer labels "Multiple choice" and "Select multiple" do not clearly describe the behavior.

This change adds an optional requirement for at least one selection while keeping existing studies compatible. Choice questions without the new setting remain optional.

Changes:
- Add the backward-compatible `selectionRequired` choice-question setting, defaulting to `false` when absent.
- Add a **Require an answer** switch when **Allow multiple selections** is enabled in the designer.
- Rename the designer question type from **Multiple choice** to **Choice** and clarify the multiple-selection option.
- Label participant-facing multi-select questions as optional or required.
- Disable empty confirmation for required questions and prevent participants from removing the final selected answer.
- Add serialization and widget regression tests.

## Visuals

Screenshot 2026-07-17 at 20 29 09
Screenshot 2026-07-17 at 20 29 58

## Testing Steps

1. Open the designer and add or edit a **Choice** question.
2. Enable **Allow multiple selections** and verify that **Require an answer** appears.
3. Save and reopen the question to verify that the requirement persists.
4. In the participant app, verify that an optional question can submit no selections.
5. Enable **Require an answer** and verify that confirmation stays disabled until the participant selects an answer.
6. After confirming, verify that the participant cannot remove the final selected answer.

Automated verification:
- `scripts/pre-commit-check`
- `cd core && fvm dart test test/questionnaire/choice_question_test.dart`
- `cd app && fvm flutter test test/widgets/questionnaire/questions/choice_question_widget_test.dart`

### PR Checklist
- [x] Formatting passes via `scripts/pre-commit-check`
- [x] Dart and Flutter analyzers pass via `scripts/pre-commit-check`
- [x] Screenshot or video attached


mohiuddinshahrukh:
praise: the feature itself works
issue: The feature cannot be accurately tested due to other bugs.
The choice requires me to enter an option even if no answer is required - From looking at this screen: 
image
Require an answer means an answer **must** be selected where as allow multiple selections would mean the user can select more than one answers - but this doesn't mean that the user **must** give an answer. Therefore either the functionality is incorrect or it requires some UI level clarification. 

Imagine I have 2 options (A and B) and It is required to select one of the two:
If I select A and B then try to de-select both - I cannot do this - One options always remains selected and this is a bit annoying also - I will share a video of this on slack. This should allow the user to deselect all option if they want - its not like they can go to the next page without selecting something 🥇

johannesvedder:
- Renamed the Designer setting: Require an answer -> Require at least one selection
- Removed the guard that prevented clearing the final selected option.
  - Required multi-choice questions can now be empty while being edited.
  - Confirm selection remains disabled until at least one option is selected.

mohiuddinshahrukh (approved):

PR #912, opened 2026-08-15
fix(designer): center studies empty state

mohiuddinshahrukh:
## Description
Centers the dashboard empty state more naturally within the visible content area when there are no studies to show.

Changes:
- compute a bounded empty-state height from the current viewport
- let the existing `EmptyBody` center itself vertically within that available space
- avoid the earlier unbounded flex layout issue in the dashboard content area

## Visuals
- Static UI change; screenshot attachment required before merge.

## Testing Steps
1. Run `rtk fvm exec melos run qualitycheck` from repo root.
2. Open the Designer dashboard with no owned studies.
3. Verify the `You don't have any studies yet` empty state appears centered more naturally in the page content area.
4. Verify the dashboard still loads normally and no layout/assertion errors appear.

### PR Checklist
- [x] `fvm exec melos qualitycheck` passes
- [x] Screenshot or video attached, or this item removed for non-visual changes
- [ ] Description links related issues, or this item removed when no issue exists

image

### DIFF
--- a/designer_v2/lib/features/dashboard/dashboard_page.dart
+++ b/designer_v2/lib/features/dashboard/dashboard_page.dart
@@ -4,6 +4,7 @@ import 'package:intl/intl.dart';
 import 'package:studyu_core/core.dart';
 import 'package:studyu_designer_v2/common_views/async_value_widget.dart';
 import 'package:studyu_designer_v2/common_views/empty_body.dart';
+import 'package:studyu_designer_v2/common_views/layout_two_column.dart';
 import 'package:studyu_designer_v2/common_views/primary_button.dart';
 import 'package:studyu_designer_v2/common_views/search.dart';
 import 'package:studyu_designer_v2/features/dashboard/dashboard_controller.dart';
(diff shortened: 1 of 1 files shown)


ibrahimozkn (changes requested):
issue (blocking): subtract the content padding only once

`EdgeInsets.vertical` already returns the sum of the top and bottom padding. `TwoColumnLayout.defaultContentPadding.vertical` is `64`, but multiplying it by `2` makes `contentPadding` equal `128`. This reduces the empty-state area by an extra 64 pixels and places its center about 32 pixels above the intended center.

Use:

```dart
final contentPadding = TwoColumnLayout.defaultContentPadding.vertical;
```

johannesvedder (commented):
`suggestion` (non-blocking): duplicated magic numbers will silently drift
`designer_v2/lib/features/dashboard/dashboard_page.dart:536-538` — `headerHeight = 36.0` duplicates the `SizedBox(height: 36.0)` at `:216`, `spacingBelowHeader = 24.0` duplicates `:528`, and `activeFilterSpacing = 64.0` approximates the `SizedBox(height: 16.0)` at `:506` plus chip height. Extract shared named constants so a spacing tweak can't desync the centering math.

`issue` (non-blocking): compact mode overestimates the remaining area by ~56px
`designer_v2/lib/features/dashboard/dashboard_page.dart:533` — `MediaQuery.sizeOf(context).height` is the full window height; on compact screens (<900, `DashboardScaffold` shows an `AppBar`) that includes the app bar area, which is never subtracted. Result: the empty state sits ~28px below true center and adds a small phantom scroll region. Deriving the height from a `LayoutBuilder` over the content column would fix compact mode and remove all the constants.

`nitpick` (non-blocking): clamp `minHeight` at zero
`designer_v2/lib/features/dashboard/dashboard_page.dart:543-548` — `emptyStateMinHeight` goes negative when the viewport is under 124px (or 188px with an active filter); debug builds then throw "BoxConstraints has a negative minimum height." from `debugAssertIsValid`. Cheap fix: `max(0.0, ...)`.

`nitpick` (non-blocking): `64.0` filter-row estimate
`designer_v2/lib/features/dashboard/dashboard_page.dart:538-542` — the actual row is `SizedBox(height: 16.0)` + a `Wrap` of 32px chips ≈ 48px, and the `Wrap` grows when chips wrap to multiple lines, so the fixed estimate diverges as filters accumulate.

`note` (non-blocking): the `studies_not_found` (query) branch lost its previous `EdgeInsets.only(top: 24.0)` and now centers like the primary empty state. Looks consistent with the goal and screenshot, just flagging it as a deliberate visual change.

`question` (non-blocking): the description mentions "the earlier unbounded flex layout issue", but the previous `EmptyBody` (`Column`, `MainAxisSize.max`, no flex children) inside the scroll view would not have thrown the unbounded-flex assertion — the change fixes centering, not an assertion. Maybe just reword.

johannesvedder:
solved in https://github.com/hpi-studyu/studyu/pull/840

PR #906, opened 2026-07-30
fix(app): harden degraded participant recovery flows

johannesvedder:
## Description

This PR hardens participant-app behavior when backend connectivity degrades, when cached participant state must be restored, and when a participant is removed or switched across backends.

Main changes:

- initializes Supabase with a fallback URL when all configured backend probes fail, so degraded startup can continue into cache/offline recovery instead of crashing with an uninitialized client
- validates stored participant session data against the active backend before reuse and clears stored participant credentials only on explicit `invalid_credentials`
- preserves stored participant credentials during backend outages and device-offline transport failures instead of treating all failures as stale credentials
- adds a shared participant-app connection-state model and persistent localized degraded-connection banner for:
  - device offline
  - backend unavailable while cached data remains usable
- routes degraded startup cache-missing and cache-corrupt failures through explicit recoverable app-error flows instead of endless loading or generic fall-through behavior
- restores cached participant state during degraded startup without creating replacement participant accounts during outages
- keeps cached study configuration, schedule, task state, and participant progress available during degraded restart flows
- hardens reconnect synchronization so cached progress merges without duplicate completions or lost progress, and retries partial failures safely
- scopes pending offline media uploads to the matching `studyId` and `userId` so reconnect cannot upload queued work into the wrong participant or backend after a switch/removal flow
- defines and implements removal policy for deleted/removed participants:
  - retain cached study data for degraded-connectivity usage
  - purge study-local cached subject state, selected subject reference, scoped pending uploads, and study-scoped credentials when participant is removed or deleted
- prevents deleted/removed participants from regaining access to stale deleted study data through cached local state or later reconnect synchronization

This is intentionally a broader stabilization PR because the original backend-switch failure exposed multiple gaps across startup, degraded operation, synchronization, and removed-subject cleanup.

## Visuals

Attach before merge:
- degraded startup with backend-unavailable banner
- degraded cached dashboard / task flow
- recovery after backend restoration

## Testing Steps

1. Run `cd flutter_common && rtk fvm flutter test test/init_test.dart test/connection_status_test.dart test/env_loader_test.dart test/retry_future_builder_test.dart`.
2. Run `cd core && rtk fvm flutter test test/model/app_config/app_config_test.dart test/model/study_subject/study_subject_cache_test.dart`.
[...]

PR #936, opened 2026-08-26
STUDYU-95: fix(app): persist and sync tasks offline

johannesvedder:
## Description

Related Jira issue: [STUDYU-95](https://studyu.atlassian.net/browse/STUDYU-95)

Split from #906 and stacked on cache recovery.

**Stack 3 of 3** · Previous: #935

Completing a task while the backend was unavailable could leave the UI loading indefinitely. Progress could be lost, duplicated, overwritten by synchronization, or restored after participant cleanup. Full reset and opt-out flows also had races with task saves, authentication recovery, and secure-storage persistence.

This PR makes offline completion durable and synchronization safe:

- completes tasks locally while degraded and persists progress before leaving the task screen;
- keeps a persistent cache Retry without rerunning or duplicating task completion;
- assigns stable completion identity for idempotent remote retry;
- serializes cache reads, writes, revisions, and deletion against synchronization;
- preserves concurrent local progress and retries failed merges safely;
- coordinates task mutations, authentication recovery, opt-out, study switching, and full reset through destructive-operation barriers;
- prevents stale subject/session data from being resurrected after cleanup;
- resumes automatic synchronization after failed deletion even if best-effort cache recovery fails.

This is Stack 3 of 3. Base: `fix/pr906-cache-recovery`.

## Testing Steps

1. Open a participant task while online, then make the backend unavailable.
2. Complete the task and verify the screen returns promptly instead of loading indefinitely.
3. Reload while offline and verify completion remains visible from cache.
4. Restore the backend and verify progress synchronizes exactly once.
5. Exercise opt-out/full reset while synchronization or authentication recovery is active and verify cleared data is not restored.
6. Automated verification completed:
   - `cd app && rtk fvm flutter test test/app_state_test.dart test/cache_synchronization_test.dart test/offline_task_persistence_test.dart test/study_local_cleanup_test.dart`
   - `rtk scripts/pre-commit-check`
   - `git diff --check`

### PR Checklist
- [x] Formatting passes via `scripts/pre-commit-check`
- [x] Dart and Flutter analyzers pass via `scripts/pre-commit-check`
- [ ] Screenshot or video attached
- [x] Description links related work (#906)

[STUDYU-95]: https://studyu.atlassian.net/browse/STUDYU-95?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/lib/models/app_state.dart
+++ b/app/lib/models/app_state.dart
@@ -29,7 +29,9 @@ class AppState with ChangeNotifier {
   late final VoidCallback _connectionStatusListener;
   AppConnectionStatus _connectionStatus = appConnectionStatusController.status;
   Timer? _activeSubjectSyncRetryTimer;
-  bool _activeSubjectSyncInFlight = false;
+  Future<void>? _activeSubjectSyncFuture;
+  bool _activeSubjectSyncStopped = false;
+  bool _activeSubjectSyncPending = false;
   StreamSubscription<StudySubject>? _activeSubjectCacheSubscription;
   StudySubject? _activeSubjectCacheSource;
 
@@ -42,6 +44,9 @@ class AppState with ChangeNotifier {
 
   @visibleForTesting
   Future<bool> Function()? debugRestoreParticipantSessionForSync;
+
+  @visibleForTesting
+  bool Function()? debugHasParticipantSessionForSync;
 
   bool get hasPendingDeepLink =>
       pendingDeepLinkStudyId != null || pendingDeepLinkInviteCode != null;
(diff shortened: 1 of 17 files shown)


johannesvedder:
1. We want to target advanced offline functionality. For worst case scenarious it should be enough to sync once for joining a study and second time after study is done to sync results to server. Edge cases
  - what happens if user has missed last day of the study and study is already over will it still sync the results
  - will fitbit data will be retrieved even if study is over?
  - what happens if user started a new study after completing one. Will the previous study will be synchronized as well?
  - What happens in case of synchronization issues?

